PR #7, opened 2018-03-18
fix: allow unpack to work when length is exactly the max field bit size

aaronwinter:
Summary: The `unpack` function defined in `snark0.ml` should work when the length of the argument passed is equal to `Field.size_in_bits`.

This PR also fixes the merkle_update proof example.

### DIFF
--- a/src/snark0.ml
+++ b/src/snark0.ml
@@ -1328,7 +1328,7 @@ module Checked = struct
   ;;
 
   let unpack v ~length =
-    assert (length < Field.size_in_bits);
+    assert (length <= Field.size_in_bits);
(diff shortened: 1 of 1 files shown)


imeckler:
Hi @aaronwinter - the change you propose is problematic. The issue is that when the size of the bit string is `>=` to that of the field size in bits, there may be multiple unpackings. This is addressed somewhat in the recently pushed changes, where there is now `choose_preimage` which doesn't care about the length and `unpack_full` which constrains the output unpacking to be (as a bit string) smaller than the size of the field, which addresses the "multiple unpackings" issue.

Issue #18, opened 2018-08-10
Build requires non-public dependencies?


gtank:
As of commit fad2a6d32ce8d5314c6daafa699f80985b865719 (the one helpfully titled "cb6a5c67d8805288f33f8eb5e6b27f3a13ee7d5f" :smiley:) I still can't build snarky. The errors are as follows:

```[ERROR] The compilation of snarky failed at "jbuilder build -p snarky -j 4".                                                                                                                          
                                                                                                                                                                                                      
#=== ERROR while installing snarky.~unknown ===================================#                                                                                                                      
# opam-version 1.2.2                                                                                                                                                                                  
# os           linux                                                                                                                                                                                  
# command      jbuilder build -p snarky -j 4                                                                                                                                                          
# path         /home/snarky/.opam/system/build/snarky.~unknown                                                                                                                                        
# compiler     system (4.05.0)                                                                                                                                                                        
# exit-code    1                                                                                                                                                                                      
# env-file     /home/snarky/.opam/system/build/snarky.~unknown/snarky-19-186c2e.env                                                                                                                   
# stdout-file  /home/snarky/.opam/system/build/snarky.~unknown/snarky-19-186c2e.out                                                                                                                   
# stderr-file  /home/snarky/.opam/system/build/snarky.~unknown/snarky-19-186c2e.err                                                                                                                   
### stderr ###                                                                                                                                                                                        
# [...]                                                                                                                                                                                               
# x - caml_mnt6_specific.cpp.o                                                                                                                                                                        
# x - pedersen.cpp.o                                                                                                                                                                                  
# x - tbcs.cpp.o                                                                                                                                                                                      
# x - memory_store_trace.cpp.o                                                                                                                                                                        
# x - ra_memory.cpp.o                                                                                                                                                                                 
# x - fooram_aux.cpp.o                                                                                                                                                                                
# x - tinyram_aux.cpp.o                                                                                                                                                                               
# ~/.opam/system/build/snarky.~unknown/_build/default/src/camlsnark_c/libsnark-caml/build ~/.opam/                                                                                                    
system/build/snarky.~unknown/_build/default/src/camlsnark_c/libsnark-caml ~/.opam/system/build/sna                                                                                                    
rky.~unknown/_build/default/src/camlsnark_c                                                                                                                                                           
# ~/.opam/system/build/snarky.~unknown/_build/default/src/camlsnark_c/libsnark-caml ~/.opam/system                                                                                                    
[...]

PR #19, opened 2018-08-11
Add missing deps

bkase:
As noted in #18 , I forgot to include two libraries that we made dependencies of snarky since the last update.

I included them here (`interval_union` and `bitstring_lib`)

I included the Dockerfile @gtank created and modified it to include the two libraries, but I'm getting a strange error in the docker container when I try to build it.

I spent a bit tonight trying to get the Dockerfile to build but I'm not super familiar with opam pinning, just local dune builds. I'm getting a weird error what seems like the end of the build and seems unrelated to the actual ocaml code.

Can anyone help me get this work? cc @nholland94

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,6 +20,7 @@ RUN apt-get -q update && \
         ocaml \
         opam \
         nano \
+        pkg-config \
         python-markdown && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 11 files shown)


gtank:
@bkase I had a few minutes to look at this today; check out #20 and see if it works for you too!

bkase:
Thanks @gtank !

PR #37, opened 2019-01-18
Tweak formatting

mrmr1993:


bkase:
We should get ocamlformat here

PR #48, opened 2019-01-23
Fix merkle_update example

### LINKED ISSUE #41: Merkle_update example fails
Please re-enable in the `make examples` target when this is fixed

mrmr1993:
This PR makes the length checks non-strict when converting from bits.

This fixes the off-by-one error that `examples/merkle_update` was hitting: `Knapsack.Checked.hash_to_bits` would try and unpack a field element into `Field.size_in_bits` field bits and then immediately hit the assertion in `unpack`.

This fixes #41.


mrmr1993:
@bkase I spoke to @imeckler about this. I think it's right based on the C code and other similar users in our code, but he said he'd have a look.

mrmr1993:
Updated to use `choose_preimage_var` instead of `unpack` in knapsack hash.

PR #53, opened 2019-02-04
Dedicated language for snarky

mrmr1993:
This is the first stage in implementing a dedicated language for Snarky with ReasonML-style syntax (current progress [here](https://github.com/o1-labs/snarky/tree/snarky-dsl)). Running the program generates an OCaml AST.

This version implements:
* `let` bindings
* function declarations with `fun`
* function application
* ~~typechecking~~
* type variables and arrows
* type names (compared by string)
* integers, bound to the `int` type

Quirks:
* non-user-named type variables are given the name `aID` where `ID` is the numeric id of the type, for debugging purposes
* the parser has some ambiguous cases (that Menhir complains about). These are fixed in a later version.
* ~~the `.ast` test output is extremely brittle, and not really useful for tests. This is replaced with more helpful output (`.mli`-style environment type information) in a later version.~~ **Removed**
* ~~some of the type variables are not fully unified. This is fixed in the lastest version.~~ **Removed**
* ~~typechecker errors aren't fatal. This is fixed in a later version.~~ **Removed**


mrmr1993:
@bkase I've pulled everything but the basics out of this; now it's only the parser and OCaml translation layer. `dune exec meja/meja.exe -- test.meja` to run the tests.

meja/test.meja
```diff
@@ -0,0 +1,20 @@
...
+
+let c = {x; y;};
+
+let d = fun (x) => { x; };
+
+let e = let e = let e = x; e; e;
+
+let f =
+  let f = fun (x) => { x; };
+  let g = (d f) f;
```
bkase: you don't have to fix it in this PR, but can you make function application "reason-style" aka: `d(f)(f)` instead of `(d f) f`.

Reason just uses an extra pair of parens for representing tuples. So `a(x, y)` is the curried version and `a((x, y))` is the tupled one.

meja.opam
```diff
@@ -0,0 +1,6 @@
+opam-version: "1.2"
```
bkase: why is it called `meja`?

meja/test.meja
mrmr1993: D'oh! That actually makes some of the parsing simpler for some later things, I can't believe I missed that. I'll drop that in the next PR.

mrmr1993:
> `let%expect` tests

In later versions, I've been using a bash script that tracks a couple of outputs, the exit code and the output on stderr, so that we can also test the error messages, output, etc. I've included it here and put it on CI; should I be writing `let%expect` tests as well/instead?

PR #54, opened 2019-02-05
Meja basic

mrmr1993:
This PR:
* makes sure that errors are logged to the given file with the `--stderr` argument
* adds the `pattern` type
* uses `pattern` to simplifiy the representation of functions, bringing it closer to the OCaml ast
* adds the arrow (`->`) type
* adds a typechecker
* adds a test for a typecheck failure

This PR **does not**:
* get rid of the `fun` keyword for function expressions
  - this makes the grammar non-LR(1), since it has to look ahead to distinguish  `(x) (x, y, z)` function application from `(x) => {...}` function expressions, and so the parser always chokes on one of the 2 cases


johnwu93:
Why is this called Meja.

meja/typechecker.ml
```diff
@@ -0,0 +1,110 @@
+open Core_kernel
```
bkase: Did you write this using typing rules written down somewhere? If so, it would help with reviewing if you could share that

meja/parsetypes.ml
```diff
@@ -1,15 +1,31 @@
 type str = string Location.loc
 
-type pattern = PVariable of str
+type type_expr = {mutable desc: type_desc; id: int}
```
bkase: It would be nice to avoid mutability if possible -- can you speak a bit about why you reached for it here?

mrmr1993: It's possible, I just leant on mutability here to make the typechecking code cleaner, simpler, and faster.

I'll rework it to avoid mutability.

meja/typechecker.ml
mrmr1993: I started with the bidirectional typechecker from the [paper we discussed](https://www.cl.cam.ac.uk/~nk480/bidir.pdf), and then wrote `check_type` below to do the actual checking and unifying to full the gap when I realised it wasn't a good fit.

mrmr1993:
> Why is this called Meja.

It needed a name, so I gave it one :man_shrugging:

meja/parsetypes.ml
imeckler: I don't have a huge problem with mutability in the type checker for what it's worth if it makes the code better.

mrmr1993:
I've rewritten the part of the typechecker that used mutability, and farmed out the state stuff to a new `Envi` module.

meja/envi.ml
```diff
@@ -0,0 +1,124 @@
+open Core_kernel
```
bkase: nit: I prefer `Env` to `Envi` everywhere, not sure if this is a preference others share

mrmr1993: We get a module conflict with OCaml's compiler libs if we do this :cry:

mrmr1993:
Done!

meja/typechecker.ml
```diff
@@ -0,0 +1,108 @@
...
+      let env = Envi.open_scope env in
+      let p_typ, env = Envi.Type.mkvar None env in
+      (* In OCaml, function arguments can't be polymorphic, so each check refines
+       them rather than instantiating the parameters. *)
+      let env = check_pattern ~add:Envi.add_in_progress env p_typ p in
+      let body_typ, env = get_expression env body in
+      let env = Envi.close_scope env in
+      Envi.Type.mk (Tarrow (p_typ, body_typ)) env
+  | Seq (e1, e2) ->
+      let _, env = get_expression env e1 in
```
cmr: I'm very curious about how this `_` will be treated as the language continue condensing. By default ocaml is going to warn if we sequence something that is non-`unit`.

meja/typechecker.ml
```diff
@@ -0,0 +1,108 @@
...
+        | Some _ ->
+            failwith
+              "Found a type variable that unifies with part of its own instance"
+        | None -> Some (Envi.Type.add_instance typ typ' env) )
+    | None -> None
+  in
+  match (typ.type_desc, ctyp.type_desc) with
+  | _, _ when Int.equal typ.type_id ctyp.type_id -> env
+  | Tvar (_, depth), Tvar (_, constr_depth) -> (
+    match without_instance typ env ~f:(fun typ -> check_type_aux typ ctyp) with
```
nholland94: This part would probably be a lot easier to read if you built a `bind_none` helper function (`val bind_none : 'a option -> f:(unit -> 'a option) -> 'a option`).

meja/parsetypes.ml
nholland94: I agree with @imeckler.

PR #56, opened 2019-02-06
Carry locations in the ast

mrmr1993:
This adds a location to all parts of the AST, so that we can have useful, located error messages later.


mrmr1993:
> What is `loc_ghost`?

It's a flag that says whether the object associated with the `Location.t` is actually in the source code at that point (as opposed to being generated by something else at that location, etc.).

PR #60, opened 2019-02-07
[Meja] Better errors

mrmr1993:
This PR
* defines some custom error types
* uses `Location.register_error_of_exn` and `Location.error_of_printer` to handle the error types and format the error messages nicely
* updates the error messages to be more informative


meja/envi.ml
```diff
@@ -117,4 +117,4 @@ let get_name name env =
   match List.find_map ~f:(Scope.get_name name) env.scope_stack with
   | Some (In_progress typ) -> (typ, env)
   | Some (Final typ) -> Type.import typ env
-  | None -> failwith "Could not find name."
+  | None -> raise (Not_found_s (Atom "Could not find name in the environment"))
```
imeckler: Maybe `sprintf` in the name?

mrmr1993: It gets immediately caught in `typecheck.ml`. I only used `Not_found_s` because `Not_found` yells at me that it's deprecated.

PR #66, opened 2019-02-08
[Meja] Add variant types

mrmr1993:
This PR
* adds variant types to the parser and environment
* moves the type declaration id table from the environment's `Scope.t` to the (unscoped) `TypeEnvi.t`
* merges error messages about unexpected structure items in pattern scopes, adds a new one for constructors
* adds some tests


meja/envi.ml
```diff
@@ -11,6 +11,7 @@ type error =
   | Wrong_number_args of string * int * int
   | Expected_type_var of type_expr
   | Lident_unhandled of string * Longident.t
+  | Constraints_not_satisfied of type_expr * type_decl
```
imeckler: what does this error mean exactly?

mrmr1993: The type constraint on a constructor of a variant type doesn't match the variant type that owns it. So like
```ocaml
type _ t = A : bool;;
```

The wording here is based off the OCaml error message in the same situation.

imeckler: Ah got it

PR #72, opened 2019-02-12
Remove `with_constraint_system`

imeckler:
This PR removes the `With_constraint_system` constructor and the `'sys` type parameter. I also deleted all the gadget code which depended on the existence of said constructor. This is fine because we wrote snarky-native equivalents of all three gadgets (verifier gadgets and SHA, which I replaced with blake2).

I think it is extremely unlikely that we will use any libsnark gadgets going forward, but it might be nice to have some kind of cross-compatibility with bellman in the future, though that would likely require a different interface anyway.


imeckler:
Oh good idea! I can do that tomorrow morning PST (or feel free to do it before then) when I am back at the computer

PR #84, opened 2019-02-15
Imperative interface

mrmr1993:
This PR creates a bare-minimum interface for running checked computations imperatively. At the moment, the interface (in `Run`) is roughly the constructors of `Checked.t`, plus a converter `of_checked` that takes an actual `Checked.t` and runs it.

This PR:
* moves the moving parts of the implementation of `Snark0.Checked.run` out into `Run`, functored over the types and utility functions it would normally have in `Snark0`.
* creates an imperative-ish API which works in a state monad containing the current `run_state`
* wraps the state-monad API in a module expression with a `run_state ref`, so that it can be passed around as an argument (of type `Run.impl`)
* adds the true imperative API, using this type of module expression as its argument (labelled `~impl`)
* exposes a `Run` module from `Snark0.Make`, which gives versions of `prove`, `verify`, etc. that operate on these imperative calls (of type `impl:('field, 'field_var) impl -> 'a`, analogous to `('a, unit, 'field, 'field_var) Checked.t`).

Quirks:
* there are some `'s prover_state` types lurking around, which decide whether the prover can have state (`type 's prover_state = 's`) in the normal case, or not (`type _ prover_state = unit`) otherwise. These prevent us from writing out the signatures even more times.
* most of the core code is now generalised over `Checked.t`, except for `Typ.t`s and a few other pain points

### DIFF
--- a/src/run.ml
+++ b/src/run.ml
@@ -20,6 +20,8 @@ module type S = sig
   val assert_square :
     ?label:string -> Field.Var.t -> Field.Var.t -> (unit, _) t
 
+  val assert_equal : ?label:string -> Field.Var.t -> Field.Var.t -> (unit, _) t
+
   val run_as_prover :
        ('a, Field.Var.t -> Field.t, 's prover_state) As_prover0.t option
     -> ('a option, 's) t
@@ -267,12 +269,14 @@ module Make
   let next_auxiliary () state = (state, !(next_auxiliary state))
 end
 
+module type S_imperative = S with type ('a, _) t = 'a and type 'a prover_state = unit
+
 module Make_imperative
     (M : Checked_intf.Backend_types) (State : sig
         include Checked_intf.Runner_state(M).S with type 'a prover_state = 'a
 
         val initial_state : unit t
-    end) : S with type ('a, _) t = 'a and type 'a prover_state = unit = struct
+    end) : S_imperative = struct
   module Stateful = Make (M) (State)
   module Field = Stateful.Field
   open Stateful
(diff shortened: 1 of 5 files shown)


src/run.ml
```diff
@@ -0,0 +1,497 @@
...
+        let%map () = such_that x in
+        x
+
+  let perform ~run req = request_witness ~run (Typ.unit ()) req
+
+  let next_auxiliary () state = (state, !(next_auxiliary state))
+end
+
+module type S_imperative = sig
+  include S with type ('a, _) t = 'a and type 'a prover_state = unit
```
imeckler: As mentioned in our side-conversation, would be good to have in here a field module that actually has operations defined on it. Otherwise you can't really write any programs. I propose we give the user a module here with the signature
```ocaml
module Field : sig
  type t
  val ( * ) : t -> t -> t
  val ( + ) : t -> t -> t
  val ( / ) : t -> t -> t
  val of_string : string -> t
  val of_int : int -> t
end
```
which will be implemented with the type t actually being Cvar.t and the operations being defined using snarky code. E.g., ( * ) will actually be

```ocaml
let ( * ) x y =
  let z = exists typ ~compute:Backend.Field.Infix.( * ) in
  assert_r1cs x y z;
  z
```
where exists and assert_r1cs are the functions in the imperative module.

We should also expose the actual backend field type so that it can be used in as_prover blocks, but I think it should be relegated to a more secondary role (say in Backend.Field or Field.Unchecked or something).

mrmr1993:
Closed in favour of #92.

PR #88, opened 2019-02-19
[Meja] Add field elements to the initial environment

mrmr1993:


meja/tests/field_elements.meja
```diff
@@ -0,0 +1,5 @@
+let zero = Snarky.Backends.Mnt4.Field.zero;
```
imeckler: I was thinking we would actually just have one "Field.t" type in scope and the user can control which backend field it actually gets compiled to at compile time. The rationale is that you usually only ever care about one field, which is whatever field is provided by the backend you're going to use.

PR #90, opened 2019-02-19
[Meja] Add support for prefix/infix operators

mrmr1993:
This PR adds support for prefix/infix operators, which will let us implement the infix `+`, `*` notations for fields, etc.


mrmr1993:
> do these support precedence, will the menhir parser do that for you? Can you add some tests around that?

@bkase yes, that's what the order of the operator precedences does. Test for this is `f` in the test file I added; the operators aren't defined, but it uses unification to check that the bracketed version matches the unbracketed version. Should I add more tests?

PR #92, opened 2019-02-20
Imperative interface v2

mrmr1993:
This PR replaces #84.

It exposes roughly the same interface, but without doing all of the module type/functor work to make progress, and goes on to expose more useful parts of the `Snark0` interface.

Currently, code should go something like this:
```ocaml
let assoc_mul x y ~intf =
  let a = Run.Field.Checked.mul ~intf x y in
  let b = Run.Field.Checked.mul ~intf y x in
  Run.Field.Checked.Assert.equal ~intf a b

module Intf = Snark.Make(Backends.Mnt4)
open Intf

let exposing = Data_spec.[Field.typ; Field.typ]
let run = Run.Perform.run ~intf:(module Intf)

let keypair = Perform.generate_keypair ~run ~exposing assoc_mul
let proof = Perform.prove ~run (Keypair.pk keypair) assoc_mul (Field.of_int 1) (Field.of_int 2)
assert (Perform.verify ~run:run (Keypair.vk keypair) assoc_mul (Field.of_int 1) (Field.of_int 2))
```


imeckler:
Looks excellent — I’d like to see if we can make it a bit more user-friendly though and I have a few concrete suggestions toward that end. (By the way, these are all things mostly descending from the original interface I wrote so I take full responsibility :)

1. Try not to bury things in modules if possible. E.g., in snark code we use Field.Checked more than field so maybe that should just be called field and the other should be Field.Unchecked or Backend.Field or something.
2. Is there some way we can use infix operators? Like we unpack the module and use the field module inside which has infix operators defined on it say?
3. I think intf isn’t quite the right name. Maybe something more like impl but even then I feel if we’re going to be typing it all the time we should call it ~m or something short. 
4. I feel the top level interface could be simpler. I guess the ideal would be if we could avoid exposing to the user functors/first class modules altogether. And what is the difference between Perform and Run.Perform?

mrmr1993:
> Is there some way we can use infix operators?

@imeckler I think the least horrible way to do this is to have an alternative `Snark0` functored module, carry the state in there, and then just let the user do:

```ocaml
module Intf = Snark_Run.Make(Backends.Mnt4)
open Intf

let checked x y z =
  Field.Infix.(x * y * z)

let exposing = Data_spec.[Field.typ; Field.typ; Field.typ]

(* generate keygens, prove, etc. *)
```

This also still gives the option of the user functoring over the type of `Snark_run.Make` if they want to switch between backends.

> I feel the top level interface could be simpler. I guess the ideal would be if we could avoid exposing to the user functors/first class modules altogether.

Is this for Meja or for Snarky in general? The style above is the cleanest I've come up with for working in OCaml, because of the namespacing concerns.

For Meja, I've been working on haskell-style typeclasses to give us implicit operator overloading (so e.g. `x * y` does the right thing if `x` and `y` are `int`, `Field.t`, `Field.Checked.t`, etc.) and other niceties, which hopefully should get rid of most of the modules and namespacing.

What do you think?

imeckler:
> @imeckler I think the least horrible way to do this is to have an alternative `Snark0` functored module, carry the state in there, and then just let the user do:

I think this would look nice, but we want code to be parametrized somehow over backend otherwise it isn't sharable like in a library. E.g., that `checked` function depends on the specifing `Impl` module. What I was thinking was more like

```
type ('a, 'b) t = T of (module Snarky.New_snark_intf with type Field.t = 'a and type Field.Unchecked.t = 'b)

let checked (T (module I)) x y z =
  let open I in
  F.(x * y * z)
```

The `T` constructor is to avoid forcing the user to give a type signature for the module.

imeckler:
> Is this for Meja or for Snarky in general? The style above is the cleanest I've come up with for working in OCaml, because of the namespacing concerns.

I was thinking for snarky in general, so that we can have something reasonably nice before the Meja stuff is done.

mrmr1993:
> I think this would look nice, but we want code to be parametrized somehow over backend otherwise it isn't sharable like in a library.

We need to carry around the state somewhere, so it's hard to properly integrate with people using the monad API, but you should be able to do

```ocaml
module T (Intf : Snarky_Run.S) = struct
  open Intf

  (* ... *)
end
```

and then include functored modules like this into any other imperative code by passing around that `Intf`, like

```ocaml
module T1 (Intf : Snarky_Run.S) = struct
  open Intf

  module T = T(Intf)

  (* ... *)
end
```

or

```ocaml
module Intf = Snark_Run.Make(Backends.Mnt4)
open Intf

module T = T(Intf)

(* ... *)
```

Can we do better than this?

mrmr1993:
@imeckler I've added an alternative model closer to what we were discussing, so we can chat about it in a bit.

mrmr1993:
@imeckler Between the current version of this PR and #97, I think we've got all the extra niceties we wanted for the imperative API.

NB: I've removed the `Run` module (with the `add_constraint ~m ...` style API) for now to keep the PR smaller, with a view to it being a separate follow-up PR when we've finalised this.

PR #100, opened 2019-02-25
[Meja] Implicit arguments

mrmr1993:
This PR
- adds the `Timplicit` type for implicit arguments (analogous to `Tarrow`)
- adds the `Unifiable` expression to hold implicit arguments, so that we can unify them
  * this uses mutability, so that we don't have to do extra passes over the AST to update it when unifying these variables
- inserts implicit arguments into bindings when the type variables it introduces are contained in those arguments

There is currently no search implemented, so implicit arguments with no type variables (e.g. `testing_show (18) : {testing_show_constr(int)} -> int` in the examples) just get the argument passed through each scope for now.

This depends on #98; without it the test will fail.

### DIFF
--- a/meja/envi.ml
+++ b/meja/envi.ml
@@ -303,7 +303,15 @@ let raw_find_type_declaration (lid : lid) env =
   | None -> raise (Error (loc, Unbound_type lid.txt))
 
 let add_module (name : str) m =
-  map_current_scope ~f:(Scope.add_module name.txt m)
+  map_current_scope ~f:(fun scope ->
+      let scope = Scope.add_module name.txt m scope in
+      { scope with
+        instances=
+          Map.merge scope.instances m.instances ~f:(fun ~key:_ data ->
+              match data with
+              | `Left x -> Some x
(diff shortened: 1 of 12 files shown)


mrmr1993:
I've now reworked this:
* implicit arguments in types are marked by a flag in `Tarrow`
* functions can be defined with explicit arguments `fun {...} => {...}` or `fun {...} (...) => {...}`
* instances are declared locally with the `instance` keyword
* instance search is performed by unification with the target type
  - failure mode is when multiple candidate instances are found. This is done at resolution time instead of declaration type, but otherwise this is equivalent to Haskell's default of `no overlapping instances`
  - if no instances are found for a fully qualified (no type variables) type, then the implicit is re-exposed on the toplevel function, in case a later callpoint has one in scope that it can pass
* instance identifiers are properly qualified across `open`s and `module`s, as you would expect.

Other small bundled fixes:
* constructors now copy the type parameters from their declaration, instead of using the originals 😅
* colon before types in function definitions (i.e. `fun (...) : .. => { ... }`)

PR #101, opened 2019-02-27
Remove continuation trick

imeckler:
This PR removes the use of continuation types in the data_spec gadt and uses h_list instead. There are definitely some downsides to this approach (namely one needs to say `H_list.` occasionally) but I think overall it has made the code simpler and make the implementation of the pretty top level api @mrmr1993 and I discussed easier.

### DIFF
--- a/examples/election/election.ml
+++ b/examples/election/election.ml
@@ -107,7 +107,7 @@ let winner ballots =
 
 let number_of_voters = 11
 
-let check_winner commitments claimed_winner =
+let check_winner H_list.([commitments; claimed_winner]) =
   let%bind w = winner commitments in
   Vote.assert_equal w claimed_winner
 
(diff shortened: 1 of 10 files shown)


mrmr1993:
> There are definitely some downsides to this approach (namely one needs to say `H_list.` occasionally) but I think overall it has made the code simpler and make the implementation of the pretty top level api @mrmr1993 and I discussed easier.

Sorry @imeckler, can I push back on this? I'm really not convinced this is a good idea.
* From my side, this doesn't make the pretty toplevel api any easier.
* From the user's side, *every user* now has to worry about `H_list.t`s.
* Users can't run tests on/debug intermediate functions easily, since they have to convert the arguments to `Hlist.t`s before they can pass them to the API.

imeckler:
Yeah, thanks for the pushback. It's not obvious to me either that this is right. 

> From my side, this doesn't make the pretty toplevel api any easier.

This I sort of disagree with, although it depends exactly on what the top-level API would be. But maybe it would be better to have a proposed top-level api first and then make the necessary changes.

> From the user's side, _every user_ now has to worry about `H_list.t`s.

This definitely.

> Users can't run tests on/debug intermediate functions easily, since they have to convert the arguments to `Hlist.t`s before they can pass them to the API.

This I disagree with. The only place you need to use H_lists is on the top-level 'main' function that you actually pass to generate_keypair. I think this is basically always a function you wouldn't be testing anyway.

PR #103, opened 2019-02-27
Prettier toplevel

mrmr1993:
This PR creates the `Proof_system` interface, currently generalised over all the frontends we expose/are planning to expose, and not exposed in the `snark0.mli` file.

Basic interface structure:
```ocaml
  (* The entry point to create a `proof_system` *)
  val create :
    ?proving_key:Proving_key.t ->
    ?verification_key:Verification_key.t ->
    ?proving_key_path:string ->
    ?verification_key_path:string ->
    ?handler:Request.Handler.t ->
    exposing:('checked, unit, 'computation, 'inputs) Data_spec.t ->
    'computation ->
    ('checked, 'inputs, 'state) proof_system

  val set_inputs : ('a, 'b, 'c) proof_system -> (unit, 'b) H_list.t -> unit
  val clear_inputs : ('a, 'b, 'c) proof_system -> unit
  val set_proving_key : ('a, 'b, 'c) proof_system -> Proving_key.t -> unit
  val proving_key : ('a, 'b, 'c) proof_system -> Proving_key.t option
  val set_verification_key : ('a, 'b, 'c) proof_system -> Verification_key.t -> unit
  val verification_key : ('a, 'b, 'c) proof_system -> Verification_key.t option
  val clear_keys : ('a, 'b, 'c) proof_system -> unit

  val set_proving_key_path : ('a, 'b, 'c) proof_system -> string -> unit
  val clear_proving_key_path : ('a, 'b, 'c) proof_system -> unit
  val set_verification_key_path : ('a, 'b, 'c) proof_system -> string -> unit
  val clear_verification_key_path : ('a, 'b, 'c) proof_system -> unit

  (* MD5 digest for the constraint system *)
  val digest : ('a, 'b, 'c) proof_system -> Md5.t

  (* Generate a new keypair for the constraint system, storing it in the [proof_system] *)
  val generate_keypair : ('a, 'c, unit) proof_system -> Keypair.t

  val run_unchecked :
    ?exposing:(unit, 'c) H_list.t ->
    ?handler:Request.Handler.t ->
    ('a, 'c, unit) proof_system -> unit -> unit * 'b

  val run_checked :
    ?exposing:(unit, 'c) H_list.t ->
    ?handler:Request.Handler.t ->
    ('a, 'c, unit) proof_system -> unit -> (unit * 'b, Error.t) result

  val check :
    ?exposing:(unit, 'c) H_list.t ->
    ?handler:Request.Handler.t ->
    ('a, 'c, unit) proof_system -> unit -> bool

  val prove :
    ?exposing:(unit, 'c) H_list.t ->
    ?proving_key:Proving_key.t ->
    ?handler:Request.Handler.t ->
    ('a, 'c, unit) proof_system -> unit -> Proof.t

  val verify :
    ?exposing:(unit, 'b) H_list.t ->
    ?verification_key:Verification_key.t ->
    ('c, 'b, 'd) proof_system -> Proof.t -> bool
end
```

@imeckler I think this roughly follows the design of the toplevel API that we discussed, what do you think?

### DIFF
--- a/src/snark0.ml
+++ b/src/snark0.ml
@@ -1263,23 +1263,23 @@ module Make_basic (Backend : Backend_intf.S) = struct
         | None ->
             failwith "run_with_input: Expected a value from run, got None."
 
-      let run_unchecked ~run ?exposing ?handler proof_system s =
+      let run_unchecked ~run ~exposing ?handler proof_system s =
         let {prover_state; _} = proof_system in
         proof_system.prover_state
         <- {prover_state with eval_constraints= false; system= None} ;
-        run_with_input ~run ?exposing ?handler proof_system s
+        run_with_input ~run ~exposing ?handler proof_system s
 
-      let run_checked' ~run ?exposing ?handler proof_system s =
+      let run_checked' ~run ~exposing ?handler proof_system s =
         let {prover_state; _} = proof_system in
         let system = R1CS_constraint_system.create () in
         proof_system.prover_state
         <- {prover_state with eval_constraints= true; system= Some system} ;
-        match run_with_input ~run ?exposing ?handler proof_system s with
+        match run_with_input ~run ~exposing ?handler proof_system s with
         | exception e -> Or_error.of_exn e
         | s, x -> Ok (s, x)
 
-      let run_checked ~run ?exposing ?handler proof_system s =
-        Or_error.map (run_checked' ~run ?exposing ?handler proof_system s)
+      let run_checked ~run ~exposing ?handler proof_system s =
+        Or_error.map (run_checked' ~run ~exposing ?handler proof_system s)
           ~f:(fun (s, x) ->
             let s', x =
               As_prover.run x
@@ -1288,8 +1288,8 @@ module Make_basic (Backend : Backend_intf.S) = struct
             in
             (s', x) )
 
-      let check ~run ?exposing ?handler proof_system s =
-        Or_error.is_ok (run_checked' ~run ?exposing ?handler proof_system s)
+      let check ~run ~exposing ?handler proof_system s =
+        Or_error.is_ok (run_checked' ~run ~exposing ?handler proof_system s)
 
       let read_proving_key proof_system =
         match proof_system.proving_key_path with
@@ -1330,42 +1330,31 @@ module Make_basic (Backend : Backend_intf.S) = struct
           ignore (run_with_input ~run ?exposing ?handler proof_system s) ) ;
         let {Checked.Runner.input; aux; _} = prover_state in
         let proving_key =
-          match proving_key with
-          | Some proving_key -> proving_key
-          | None -> (
-            match proof_system.proving_key with
-            | Some proving_key -> proving_key
-            | None -> (
-              match read_proving_key proof_system with
-              | Some proving_key -> proving_key
-              | None ->
-                  let {Keypair.pk; vk} = generate_keypair ~run proof_system in
-                  write_verification_key proof_system vk ;
-                  pk ) )
+          List.find_map_exn
+            ~f:(fun f -> f ())
+            [ (fun () -> proving_key)
+            ; (fun () -> proof_system.proving_key)
+            ; (fun () -> read_proving_key proof_system)
+            ; (fun () ->
+                let {Keypair.pk; vk} = generate_keypair ~run proof_system in
+                write_verification_key proof_system vk ;
+                Some pk ) ]
         in
         write_proving_key proof_system proving_key ;
         Proof.create proving_key ~primary:input ~auxiliary:aux
 
-      let verify ~run ?exposing ?verification_key proof_system proof =
-        ( match exposing with
-        | Some exposing -> set_inputs proof_system exposing
-        | None ->
-            if not !(proof_system.has_input) then
-              failwith
-                "Could not verify the proof; no input has been provided." ) ;
+      let verify ~run ~exposing ?verification_key proof_system proof =
+        set_inputs proof_system exposing ;
         let verification_key =
-          match verification_key with
-          | Some verification_key -> verification_key
-          | None -> (
-            match proof_system.verification_key with
-            | Some verification_key -> verification_key
-            | None -> (
-              match read_verification_key proof_system with
-              | Some verification_key -> verification_key
-              | None ->
-                  failwith
-                    "Could not verify the proof; no verification key has been \
-                     provided." ) )
+          List.find_map_exn
+            ~f:(fun f -> f ())
+            [ (fun () -> verification_key)
+            ; (fun () -> proof_system.verification_key)
+            ; (fun () -> read_verification_key proof_system)
+            ; (fun () ->
+                failwith
+                  "Could not verify the proof; no verification key has been \
+                   provided." ) ]
         in
         write_verification_key proof_system verification_key ;
         Proof.verify proof verification_key proof_system.prover_state.input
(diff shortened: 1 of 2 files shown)


imeckler:
This seems about right to me, but why the imperative setters?

mrmr1993:
@imeckler I've thrown out the imperative setters.

Other changes:
* this is now exposed in both the old interface and the imperative interface
* a few optional arguments are now labelled arguments, since we have no separate setters

AFAICT, this is ready to go on my end!

imeckler:
Very neat -- do you mind making an example that uses the API? The only other question I'd raise is should we use async or something?

src/snark0.ml
```diff
@@ -1066,6 +1066,311 @@ module Make_basic (Backend : Backend_intf.S) = struct
...
+                  let {Keypair.pk; vk} = generate_keypair ~run proof_system in
+                  write_verification_key proof_system vk ;
+                  pk ) )
+        in
+        write_proving_key proof_system proving_key ;
+        Proof.create proving_key ~primary:input ~auxiliary:aux
+
+      let verify ~run ?exposing ?verification_key proof_system proof =
+        ( match exposing with
+        | Some exposing -> set_inputs proof_system exposing
```
imeckler: Why is `exposing` optional?

src/snark_intf.ml
```diff
@@ -1157,6 +1211,56 @@ module type Run = sig
...
+  module Proof_system : sig
+    type ('a, 'inputs) t
+
+    val create :
+         ?proving_key:Proving_key.t
+      -> ?verification_key:Verification_key.t
+      -> ?proving_key_path:string
+      -> ?verification_key_path:string
+      -> ?handler:Request.Handler.t
+      -> exposing:(unit -> 'a, unit, 'computation, 'input) Data_spec.t
```
imeckler: I was thinking we should change the name of this to `public_input` to be more in sync with the usual nomenclature

src/snark0.ml
```diff
@@ -1066,6 +1066,311 @@ module Make_basic (Backend : Backend_intf.S) = struct
...
+          ( match exposing with
+          | Some exposing -> set_inputs proof_system exposing
+          | None -> () ) ;
+          let system = R1CS_constraint_system.create () in
+          proof_system.prover_state
+          <- {prover_state with eval_constraints= false; system= Some system} ;
+          ignore (run_with_input ~run ?exposing ?handler proof_system s) ) ;
+        let {Checked.Runner.input; aux; _} = prover_state in
+        let proving_key =
+          match proving_key with
```
imeckler: Consider using something else for the control flow here to avoid repetition like
```ocaml
List.find_map ~f:Lazy.force
  [ lazy proving_key
  ; lazy proof_system.proving_key
  ; lazy (read_proving_key proof_system) ]
```

src/snark0.ml
```diff
@@ -1066,6 +1066,311 @@ module Make_basic (Backend : Backend_intf.S) = struct
...
+      let v = alloc_var next_input () in
+      Field.Vector.emplace_back primary_input x ;
+      v
+
+    module Proof_system = struct
+      type ('checked, 'inputs, 's) proof_system =
+        { compute: unit -> 'checked
+        ; check_inputs: (unit, 's) Checked.t
+        ; provide_inputs: (unit, 'inputs) H_list.t -> unit
+        ; has_input: bool ref
```
imeckler: why are `has_input, has_run_with_input` refs?

mrmr1993:
> I'm a bit confused by all the mutation -- what's the need?

Basically, in the interest of not re-evaluating the constraint system over and over to get the different parts of the input, we need to keep some record of what state we're in. The obvious way to do this was mutability, since
* all the parts of the state backed by libsnark are mutable too
* it saves us from returning a `Proof_system.t` everywhere to pass around this information

 The general idea is that you can get everything with just evaluating the constraint system once, looking something like:

```ocaml
let main x y : ((Field.t, unit) As_prover.t, unit) Checked.t = ...

let proof_system = Proof_system.create ~proving_key_path:"main.pk" ~verification_key_path:"main.vk"
    ~exposing:Data_spec.[Field.typ; Field.typ] main

(* We can get the return value from [main].. *)
let x : (unit * Field.t) Or_error.t =
  Proof_system.run_checked proof_system ~exposing:H_list.[Field.one; Field.of_int 15] ()

(* ..and then re-use this computation to generate the proof. *)
let proof = Proof_system.prove proof_system ()

(* In case you wanted the MD5 digest: *)
let digest = Proof_system.digest proof_system

assert (Proof_system.verify proof_system ~exposing:H_list.[Field.one; Field.of_int 15] proof)
```

or, if we don't return a useful value from our `main`:

```ocaml
let main x y : (unit, unit) Checked.t = ...

let proof_system = Proof_system.create ~proving_key_path:"main.pk" ~verification_key_path:"main.vk"
    ~exposing:Data_spec.[Field.typ; Field.typ] main

let proof = Proof_system.prove proof_system ~exposing:H_list.[Field.one; Field.of_int 15] ()

let digest = Proof_system.digest proof_system

assert (Proof_system.verify proof_system ~exposing:H_list.[Field.one; Field.of_int 15] proof)
```

The first case is the reason why I included `exposing` as an optional argument for `prove`; if the proof system has already been run, not including it is a way for the user to say "I've already run this, just create a proof". I've now removed all the other optional arguments.

imeckler:
I'm skeptical that this is a good idea. It seems like it could easily result in a very confusing use of memory. E.g. I would expect that if I call `run_checked` it frees the auxiliary input vector allocated in the course of running it. It's super confusing to me that calling `run_checked` causes a big vector to be in memory forever.

Generally, I think we should try to have as semantically clean an interface as possible. To give an idea of what I mean by semantically clean, conceptually we can think of it in the following way:

`Proof_system.create` should essentially map a function `f : 'a -> 'b` to a record of functions
```ocaml
f_prove : 'a * state -> 'b * Proof.t
f_verify : ('a * 'b) * Proof.t -> bool
```
Now there is some other complexity caused by key-generation etc., but I think this really ought to be the "ideal" conceptual framing. The stuff I think we should use optional arguments and mutation for is the stuff that gets in the way of this conceptual framing to try an expose as close an interface to this as possible.

src/snark0.ml
```diff
@@ -1066,6 +1066,253 @@ module Make_basic (Backend : Backend_intf.S) = struct
...
+        in
+        let {Checked.Runner.input; aux; _} = state in
+        let proving_key =
+          List.find_map_exn
+            ~f:(fun f -> f ())
+            [ (fun () -> proving_key)
+            ; (fun () -> proof_system.proving_key)
+            ; (fun () -> read_proving_key proof_system)
+            ; (fun () ->
+                let {Keypair.pk; vk} = generate_keypair ~run proof_system in
```
imeckler: Should this also call `write_proving_key`?

src/snark0.ml
```diff
@@ -1066,6 +1066,253 @@ module Make_basic (Backend : Backend_intf.S) = struct
...
+            ~f:(fun f -> f ())
+            [ (fun () -> verification_key)
+            ; (fun () -> proof_system.verification_key)
+            ; (fun () -> read_verification_key proof_system)
+            ; (fun () ->
+                failwith
+                  "Could not verify the proof; no verification key has been \
+                   provided." ) ]
+        in
+        write_verification_key proof_system verification_key ;
```
imeckler: I don't think we should be writing to disk when verify gets called (I think that can happen just once when we conditionally generate new keys)

src/snark0.ml
mrmr1993: `write_proving_key` is called afterwards, so that a user-provided proving key also gets written. I can pull that out if it's not what we want.

PR #108, opened 2019-02-28
753 bit curves

imeckler:
This pulls the 753 bit curves into libsnark and into the libsnark.ml bindings. It's mostly a sed job.


daira:
Isn't MNT4-753 a bit small wrt G_T security? That's roughly equivalent to the G_T security of a 251-bit BN curve. I would think that the Kim–Barbulescu improvements to NFS cast some doubt on the long-term security of that parameter choice (even if it isn't feasibly breakable yet).

imeckler:
According to personal communications with Barbulescu, a 768 bit modulus gives ~131 bits of security (he says they were over-secured before).

PR #118, opened 2019-03-05
API fixes

mrmr1993:
This PR
* fixes the type of `run_and_check` so that it accepts a function instead of a value
* add some failure messages when code is run with the system in the wrong state, in particular
  - an error when checked code is run inside an `As_prover` block, since the constraint systems will differ between prover/non-prover runs
  - an error when checked code is called outside the APIs (`Proof_system.*` or the normal `run_and_check`, `run_unchecked`, etc.)
* tweaks the types of the `Proof_system` APIs to accept a list of `Handler.t`s and build the `Request.Handler.t` out of them internally.
  - this also makes the handlers incremental, so a `Proof_system.create` can declare some handlers, and then `Proof_system.prove` can add some more.

@imeckler I got your code up and running under this, but the constraint system is unsatisfied; is that what you would expect? Also, I'm not sure what this comment means:

> `  (* This doesn't work because the "public_input" type on proof_system needs to be polymorphic in the result type *)`


src/snark_intf.ml
```diff
@@ -549,29 +563,29 @@ let multiply3 (x : Field.Var.t) (y : Field.Var.t) (z : Field.Var.t)
 
     val run_unchecked :
          public_input:(unit, 'public_input) H_list.t
-      -> ?handler:Request.Handler.t
+      -> ?handlers:Handler.t list
```
imeckler: I think this could just be a function `Request.t -> Response.t` -- what do you think?

mrmr1993: I like the idea of it being a list, so you can pass several separately-defined handlers without having to jerry-rig your own `Request.Handler.run` to stitch them together. Switching from `Handler.t` to `Request.t -> Request.Response.t` makes sense to me though.

mrmr1993: @imeckler I've just had a proper look at this, seems like we can't make the change because we need the record field polymorphism from `Request.request`. I think the only real change we could make on this is dropping the `list` (or not).

PR #130, opened 2019-03-12
Don't reformat in the node_modules subdir

mrmr1993:
One of our `node_modules ` for the website includes a `docs.mli` file which isn't an OCaml interface file. This PR stops us from trying to reformat anything under `node_modules`.

### DIFF
--- a/app/reformat-snarky/reformat.ml
+++ b/app/reformat-snarky/reformat.ml
@@ -25,7 +25,8 @@ let main dry_run check path =
             && (not (String.is_suffix ~suffix:"stationary" path))
             && (not (String.is_suffix ~suffix:".un~" path))
             && (not (String.is_suffix ~suffix:"external" path))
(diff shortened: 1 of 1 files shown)


bkase:
Yes please add this to coda repo too

app/reformat-snarky/reformat.ml
```diff
@@ -25,7 +25,8 @@ let main dry_run check path =
             && (not (String.is_suffix ~suffix:"stationary" path))
             && (not (String.is_suffix ~suffix:".un~" path))
             && (not (String.is_suffix ~suffix:"external" path))
-            && not (String.is_suffix ~suffix:"ocamlformat" path)
+            && (not (String.is_suffix ~suffix:"ocamlformat" path))
```
psteckler: Is there some kind of mapping to be used here, instead of repeating code?

PR #132, opened 2019-03-12
Add documentation comments to Snark_intf

mrmr1993:
This adds way more extensive doc-comments to `Snark_intf.S`.

Much of this copy can be re-used for `Snark_intf.Run.S`; I will make a follow-up PR for that when we have settled on the wording here.

@rbkhmrcr could you cast your editorial eye over this please? :smile:


src/snark_intf.ml
```diff
@@ -599,7 +931,19 @@ let multiply3 (x : Field.Var.t) (y : Field.Var.t) (z : Field.Var.t)
...
+    (** Run the checked computation as the prover, generating a {!type:Proof.t}
+        that the verifier may check efficiently.
+        - The [proving_key] argument overrides the argument given to
+          {!val:create}, if any.
+        - The [handlers] argument adds handlers to those already given to
+          {!val:create}. If handlers for the same requests were provided to
+          both, the ones passed here are given priority.
+    *)
 
+    (* Verify a {!type:Proof.t} generated by a prover.
```
rbkhmrcr: Should this be moved below the function, and an extra `*` added?

PR #135, opened 2019-03-13
Turn on PERFORMANCE flag for libsnark

mrmr1993:
cc @jkrauska


enolan:
Do benchmarks improve at all?

PR #136, opened 2019-03-13
Turn back on debug logging inside PERFORMANCE

mrmr1993:


mrmr1993:
Never mind, this actually makes things slower than without `DPERFORMANCE`.

PR #141, opened 2019-03-15
Expose an optional public_output parameter to Proof_system methods

mrmr1993:
This PR adds a parameter to `Proof_system.create` and `Proof_system.verify` that handles a 'public output'.

This is implemented as an additional input that
* is not passed to the checked computation
* has no known value until the end of the checked computation
* is verified to equal the return value of the function, by a user-passed `assert_equal`

This lets us avoid precomputing the expected output before we begin the proof; see the election example for example usage.


mrmr1993:
Stale, done elsewhere

PR #145, opened 2019-03-18
Use mutable add for Cvar.eval

mrmr1993:
This speeds up `dune exec coda tra` from 51.6s to 49.3s (average of 3 runs each)


src/cvar.ml
```diff
@@ -27,14 +27,15 @@ struct
...
-    let res = ref Field.zero in
+    let open Field in
+    let res = of_int 0 in
     let rec go scale = function
-      | Constant c -> res := Field.add !res (Field.mul scale c)
-      | Var v -> res := Field.add !res (Field.mul scale (context v))
-      | Scale (s, t) -> go (Field.mul s scale) t
+      | Constant c -> res += mul scale c
+      | Var v -> res += mul scale (context v)
+      | Scale (s, t) -> go (mul s scale) t
```
imeckler: Can we use mutable multiplication for the scale too?

src/cvar.ml
```diff
@@ -27,14 +27,15 @@ struct
...
 
   let eval context t0 =
-    let res = ref Field.zero in
+    let open Field in
+    let res = of_int 0 in
     let rec go scale = function
-      | Constant c -> res := Field.add !res (Field.mul scale c)
-      | Var v -> res := Field.add !res (Field.mul scale (context v))
-      | Scale (s, t) -> go (Field.mul s scale) t
+      | Constant c -> res += mul scale c
```
imeckler: It also might be worth having one field element that we copy the constant `c` into and then do a mutable multiply with `scale`

src/cvar.ml
```diff
@@ -27,14 +27,15 @@ struct
...
+      | Var v -> res += mul scale (context v)
```
imeckler: similarly here

imeckler:
just realized that's probably not the best benchmark to do -- I would edit the transaction_snark_profiler.ml do only do the auxiliary input generation. This includes keygen and a whole bunch of other stuff that we wouldn't be improving. Also, as is this algorithm still allocates a ton so I think there is still more performance to be gained.

src/cvar.ml
mrmr1993: @imeckler Won't this lead to the same number of allocations as currently?

src/cvar.ml
mrmr1993: This would mean we need an extra allocation for each add. I'll give it a try work a benchmark, but it seems like this might slow down additions like packing/unpacking

src/cvar.ml
imeckler: I guess I was thinking just have one constant that you allocate outside the loop like
```
let c0 = Field.of_int 0 in
let rec go = ...
  | Constant c ->
    Field.copy ~src:c ~dst:c0;
    c0 *= scale;
    res += c0
```

imeckler:
Ok -- just occurred to me: what about something like this? Does this algorithm make sense? I think it should be correct...

```ocaml
    let res1 = of_int 0 in
    let res2 = of_int 0 in
    let module I = struct type t = One | Two end in
    let (:=) = failwith "TODO" in
    let rec go (i : I.t) res = function
      | Constant c -> res := c
      | Var v -> res := context v
      | Scale (s, t) ->
        go i res t;
        res *= s
      | Add (t1, t2) ->
        go One res1 t1;
        go Two res2 t2;
        let other = 
          match i with
          | One -> res2
          | Two -> res1
        in
        res += other
    in
```

mrmr1993:
@imeckler I think `res1` will get clobbered by any further additions inside the second branch of `Add`, e.g. `Add (t1, Add (t2, t3))`, which will use `res1` for `t1` and then again for `t2` before using the result.

imeckler:
> @imeckler I think `res1` will get clobbered by any further additions inside the second branch of `Add`, e.g. `Add (t1, Add (t2, t3))`, which will use `res1` for `t1` and then again for `t2` before using the result.

D'oh yeah. That makes no sense :0

imeckler:
Another thought that occurs to me is we could probably implement eval which for some important special cases that actually arise in practice uses a constant amount of memory. For example if the Cvar has the form `(a1 * x1) + b1 * ((a2 * x2) + b2 * ((a3 * x3) + b3 * ( ...` then we can evaluate in constant space, and I think this is a pretty common situation.

Either way I guess the way forward would be to have a benchmark on `eval` specifically, and a benchmark running the same number of additions and multiplications as occur in an eval, but without allocating, and see the difference in times to see how far off we are.

mrmr1993:
@imeckler I've added another commit here which improves the average time (on the same benchmark) by another second-ish to 47.8s. I'll get some more useful benchmarking together tomorrow, but an ~7% overall speedup seems pretty nice!

mrmr1993:
Updated to use mutabiity for `scale` whenever it's not going to be used somewhere else. New average runtime is 47.1s.

src/camlsnark_c/libsnark-caml/libsnark/caml/caml_mnt4753.cpp
```diff
@@ -221,6 +221,10 @@ void camlsnark_mnt4753_field_mut_sub(FieldT* x, FieldT* y) {
   *x -= *y;
 }
 
+void camlsnark_mnt4753_field_copy(FieldT* x, FieldT* y) {
+  *x = *y;
```
imeckler: I'm not sure this is doing the copying correctly. I think it might be leaking the memory pointed to by x. The right thing to do would be to copy over the underlying buffer of one to the underlying buffer of the other.

src/cvar.ml
```diff
@@ -26,15 +26,38 @@ struct
     let of_index v = Var v
   end
 
+  let scratch = Field.of_int 0
+
   let eval context t0 =
-    let res = ref Field.zero in
+    let open Field in
+    let res = of_int 0 in
+    let can_mutate_scale = ref false in
```
imeckler: Is this actually safe? Like if the scalar is Field.one, wouldn't this mutate Field.one?

mrmr1993: `scale` is an internal variable only set within this function, and `can_mutate_scale` is updated every time its status changes. So
* we call `go one t0` (ie. `scale = one`) from the outer `eval` with `can_mutate_scale = false` (to avoid an extra field element allocation on every call to `eval`)
* we set `can_mutate_scale = false` for the first branch of an `Add`, resetting it to its original value for the second, to make sure that `scale` is preserved for that second branch
* we set `can_mutate_scale = true` when we see a `Scale` with `can_mutate_scale = false`, since the `Field.mul` we do there gives us a fresh field element (and reset it after, because we know the parent context has an immutable `scale`).

mrmr1993:
Updated to use the GMP `copyi` for the field `copy ~over` function. Profile timing improvements using a auxiliary input only benchmark:
* `16.89s (avg)` 1c0e0ef093cbc83f9c3ca309a23086994d50ef43 (base commit)
  - `17.34`, `16.99`, `16.57`, `16.52`
* `16.14s (avg)` cb07941
  - `16.59`, `16.22`, `15.87`, `15.88`
* `16.21s (avg)` 025710b
  - `16.83`, `16.03`, `16.09`, `15.89`
* `15.06s (avg)` 66fcb8d
  - `15.53`, `14.75`, `15.09`, `14.88`
* `14.74s (avg)` 27eafee
  - `14.92`, `14.63`, `14.96`, `14.43`
* `14.69s (avg)` a80cb1605a76c546688b81267535665da7584597
  - `15.04`, `14.39`, `14.57`, `14.74`

PR #161, opened 2019-03-29
bg snark in snarky

imeckler:
Accidentally merged this into a parent instead of master. @bkase had previously approved


src/libsnark.ml
```diff
@@ -1465,6 +1491,143 @@ module Bn128 = Make_full (struct
...
+
+      include T
+      include Make_proof_accessors (M) (T) (G1) (G2)
+
+      let delta_prime = func "delta_prime" G2.typ G2.delete
+
+      let verify_components ~a ~b ~c ~delta_prime ~z ~y_s key input =
+        let stub =
+          foreign
+            (func_name "verify_components")
```
mrmr1993: Should `Mnt4` and `Mnt6` have this function too? I can find it in `Bn128`, but the others have a `verify` function instead.

imeckler: good catch, thank you!

PR #163, opened 2019-03-31
Feature/prove aux args

imeckler:
This PR makes it possible for the `prove` functions to take additional arguments for certain backends. This is motivated by the implementation of the BG snark, which is a signature-of-knowledge, meaning the "prove" function can take a "message" for which the proof is like a digital-signature of that message (with respect to the fact that the signer was able to provide a witness for the corresponding snarky program).

I had to turn off error-warning 16 (which says an optional argument can't be erased) because the OCaml compiler doesn't realize it gets erased for some instantiation of the functor. There is only one line which triggers this warning though and it would be nice not to turn it off in general. Is there a way to disable a warning for a particular instance of that warning?

### DIFF
--- a/src/backend_intf.ml
+++ b/src/backend_intf.ml
@@ -61,12 +61,17 @@ module type S = sig
   end
 
   module Proof : sig
+    type _ with_create_args
+
+    val proof_system_type : ('a, 'a with_create_args) Proof_system_type.t
+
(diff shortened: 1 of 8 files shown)


imeckler:
I guess I didn’t want to pollute the interface with something irrelevant. Maybe the work around is sufficiently complicated that it is worth it though.

PR #164, opened 2019-04-01
Add a Proof.message type and an optional message argument to prove/verify

mrmr1993:
This PR adds a `message` type and an optional `message` argument to `Proof.create`, `prove` and `verify`, to support messages in BG snarks.

This is an alternative to PR #163.

### DIFF
--- a/src/backend_intf.ml
+++ b/src/backend_intf.ml
@@ -61,14 +61,21 @@ module type S = sig
   end
 
   module Proof : sig
+    type message
+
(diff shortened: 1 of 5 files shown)


imeckler:
Yeah let's do this and maybe reconsider the more complicated thing if there ends up being another 'type' of proof system besides signature of knowledge.

mrmr1993:
> maybe reconsider the more complicated thing if there ends up being another 'type' of proof system

Sounds good to me!

PR #169, opened 2019-04-04
Improve proving key serialization

imeckler:
This PR makes it so that the proving key does not contain its constraint system when it gets serialized. This makes it a lot smaller and was necessary to get the medium-curves to work in a simple way since previously the proving keys were too big for bin_prot.

It's pretty hacky and will maybe have to get unwound if/when we generalize the backend interface to support non-libsnark backends, but for now it was the easiest thing I could think of.

The proving key needs the constraints to be set correctly on its constraint system, so we set them as we generate the auxiliary input for the first time on a given proving key.


mrmr1993:
I've deleted the `.orig` files. Merging.

imeckler:
Thanks @mrmr1993 my bad checking those in!

PR #185, opened 2019-04-13
upgrade the toolchain + opam deps

cmr:
not terribly much snarky impact, modulo the ocamlformat version bump.


cmr:
the `opam-pin` job is failing because it's trying to build snarky master with the new container, which doesn't have compatible library versions.

PR #190, opened 2019-04-16
[Meja] Generate typs for records

mrmr1993:
This PR adds the code generation for `Typ.t`s for record types.

The generation is split into three cases:
* types which refer to a built-in type that we don't have `Typ.t`s for, such as `int`
  - these types don't get a generated `Typ.t`
* types which don't refer to any named types, only type variables
  - the type declaration is left as is, and a `Typ.t` is generated
* types which refer to one or more named types
  - the type declaration is replaced with a polymorphic declaration (`poly`)
  - the original type is replaced with an instantiation of `poly` with the appropriate parameters
  - a `var` type is generated with an instantiation of `poly` with the appropriate parameters
  - a `Typ.t` is generated, with its `'value` type parameter restricted to the original type

At present, the types and the `Typ.t` aren't loaded into the environment. When #184 is merged, the appropriate types can be loaded into scope, which will let the type definitions and the instances typecheck.

In the snarky environment that we will load post-#184, `Typ.store`/`read`/`alloc`/`check` are marked with `Typ.t` as an implicit parameter. With that, all the `Typ.t` machinery should be automatic, and users won't need to define or use `Typ.t`s maually.


meja/typechecker.ml
```diff
@@ -865,6 +865,8 @@ let type_extension ~loc variant ctors env =
   in
   (env, variant, ctors)
 
+let in_decl = ref false
```
imeckler: maybe stick this ref inside of check_statement like

```ocaml
let check_statement =
  let in_decl = ref false in
  let rec check_statement env stmt = ... in
  check_statement

File: meja/codegen.ml [hunk 2]
```diff
@@ -0,0 +1,236 @@
...
+                let name = variant.var_ident in
+                let variant =
+                  { variant with
+                    var_ident=
+                      Location.mkloc (var_type_lident name.txt) name.loc }
+                in
+                Envi.Type.mk ~loc (Tctor variant) env :: l )
+          in
+          mk_decl params
+          (*let var_decl_fields =
```
imeckler: remove?

meja/typechecker.ml
mrmr1993: The `Codegen.typ_of_decl` call in `check_statement` calls back into `check_statement`, we need it outside to stop the `poly` and `var` declarations from getting their own `poly` and `var` declarations, and so on.

meja/tests/match.ml
```diff
@@ -9,7 +9,34 @@ type t = {a: int; b: bool}
 
 type u = {f: int -> int}
 
-type ('a, 'b) v = {x: 'a; y: 'b; g: 'a -> 'b}
+include struct
+  type ('a, 'b) v = {x: 'a; y: 'b; g: 'a -> 'b}
```
imeckler: For arrow types shouldn't we refuse to generate a typ?

PR #193, opened 2019-04-18
Beefed-up imperative interface

mrmr1993:
This PR extends the imperative API interface so that it exposes all of the same functionality as the monadic interface.


src/number.ml
```diff
@@ -208,3 +208,206 @@ module Make (Impl : Snark_intf.Basic) = struct
...
+      ; var: Field.t
+      ; bits: Boolean.var list option }
+
+    let two_to_the n =
+      let rec go acc i =
+        if i <= 0 then acc else go (Field.Constant.add acc acc) (i - 1)
+      in
+      go Field.Constant.one n
+
+    let to_bits {var; bits; upper_bound; lower_bound= _} =
```
imeckler: Can we use the monadic versions to implement all these? It's a bit sad to duplicate all this code.

mrmr1993: Done

PR #211, opened 2019-04-29
Let libsnark handle writes that may be called from multiple threads

mrmr1993:
This fixes an occasional race condition in logging through the coda logger


src/libsnark.ml
```diff
@@ -29,9 +29,11 @@ let set_printing_file =
   foreign "camlsnark_set_printing_file" (string @-> returning void)
 
 module Print_func = struct
-  let print = ref (fun _ -> ())
+  let print = ref (fun str -> str)
```
imeckler: What's this ref about?

mrmr1993: Added some (hopefully) clarifying comments.

The `ref` is to hold the user-provided printing function. We can't hold it on the C++ side because it might get relocated by OCaml's GC, so we go through `dispatch` instead because we can guarantee that will be static in memory.

PR #218, opened 2019-05-02
Add a SNARKY_PERFORMANCE environment variable to toggle the libsnark flag

mrmr1993:
This PR
* makes dune aware of the `SNARKY_PERFORMANCE` environment variable
* changes the default value of the libsnark `PERFORMANCE` flag back to `OFF`
* uses the environment variable to toggle the flag

Note that this flag may be set for certain dune profiles in the `dune-workspace` file, so we can have it always enabled for some builds if we want.

@imeckler coda builds after a `make clean` with `PERFORMANCE=ON` take nearly double the time that they did before, so we should probably do this before we merge the latest snarky there.


imeckler:
Seems good to me -- @nholland94 @cmr @jkrauska do any of you have an opinion based on how things went with the consensus mechanism environment variable? I think this is probably pretty different but just wanted to check

PR #224, opened 2019-05-06
[Meja] Separate parsed types from typechecker types

mrmr1993:
This PR
* separates the type representation generated from the parser and the representation used by the typechecker
* adds the `Typet` module to convert parsed types into typechecker types, adding them to the environment
* adds the `Untype_ast` module to convert typechecker types back into parsed types
* fixes the errors where parsed types were introduced to the typechecker without typechecking first
  - most notably, this fixes some bugs in the implicit type arguments code
  - this also changes the behaviour of `switch` statements slightly (for the better)
* updates the test output with the new behaviour
* removes a test that should now fail (`int` should no longer unify with `field`)

There is some follow-up tidying up to do (e.g. giving the different type kinds different constructor names, removing some code in `Envi` that is unused, etc.) but this PR is already quite large.


mrmr1993:
Rebased onto master

meja/src/ast_build.ml
```diff
@@ -46,13 +47,12 @@ module Type = struct
   let variant ?loc ?(params = []) ?(implicits = []) ident =
     { var_ident= Loc.mk ident ?loc
     ; var_params= params
-    ; var_implicit_params= implicits
-    ; var_decl_id= -1 }
+    ; var_implicit_params= implicits }
 
-  let none ?loc ?(explicit = Explicit) () = mk ?loc (Tvar (None, -1, explicit))
```
imeckler: :+1:

PR #225, opened 2019-05-07
As_prover lenses and imperative interface state

mrmr1993:
This PR
* adds lenses as the `Lens` module, effectively a getter/setter pair
* adds mappers for `As_prover.t` and `Checked.t` to modify the state argument based on the lens
* adds non-unit state to the imperative API
  - this is a breaking API change: `prove`/`run_checked`/`run_unchecked`/`check` and creating the imperative API module itself now all require 1 additional parameter
  - this change brings the changed APIs in line with the monadic equivalents
* uses the imperative API's state and the lens interface to create imperative versions of `Knapsack` and `Merkle_tree`

This brings the imperative and monadic APIs up to equivalence in the libraries bundled with Snarky: any functionality that you can access through the monadic API is now available in the imperative API.


src/merkle_tree.ml
```diff
@@ -518,3 +518,104 @@ struct
...
+
+          val prover_state : Impl.prover_state
+          (** The prover state to run the checked computations above with.
+              This state will *always* be passed to the above unchanged.
+
+              NOTE: This is equivalent to the condition on the monadic
+                    interface that the computations are not constrained in
+                    their prover-state type: the type is abstract from the
+                    perspective of the functions, and so they cannot have any
+                    affect on the state.
```
cmr: s/affect/effect

src/merkle_tree.ml
```diff
@@ -518,3 +518,104 @@ struct
...
+
+        val prover_state : Impl.prover_state
+        (** The prover state to run the checked computations above with.
+              This state will *always* be passed to the above unchanged.
+
+              NOTE: This is equivalent to the condition on the monadic
+                    interface that the computations are not constrained in
+                    their prover-state type: the type is abstract from the
+                    perspective of the functions, and so they cannot have any
+                    affect on the state.
```
cmr: s/affect/effect

PR #226, opened 2019-05-07
Run checked functions as the prover/outside checked computations

mrmr1993:
This PR
* replaces writes (via `exists`) made outside the checked part of checked computations with fake writes, returning a constant `Cvar.t` containing the value
* disables constraint additions (via `add_constraint`) outside the checked part of checked computations
* sets the state to be in `as_prover` mode by default, so checked computations can be run at toplevel and between proofs/checks/etc.
* adds an example/test to check that this is working correctly, set to run on the CI during `make examples`
* tweaks the semantics of the `is_running` flag to catch errors of the form below
  - the typechecker can't catch this because it has the right type (`input1 -> input2 -> ... -> inputn -> unit -> output`)
  - this also clarifies the error message and makes sure the error is thrown at a useful point (ie. the first invalid call)

```ocaml
let checked1 x y () = Field.(x * y)

let checked x y z =
  let a = Field.(x * y) in (* error at this call to Field.( * ) *)
  checked1 z a

let checked x y z () =
  let a = Field.(x * y) in (* no error: the whole computation is after the () argument *)
  checked1 z a ()
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,10 @@ publish-website : website
 
 examples :
 	dune exec --root=. ./examples/election/election_main.exe
-	# TODO: Re-enable when fixed, see #41
 	dune exec --root=. ./examples/merkle_update/merkle_update.exe
 	# tutorial.exe intentionally is unimplemented, but it should still compile
 	dune build --root=. ./examples/tutorial/tutorial.exe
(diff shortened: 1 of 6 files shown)


mrmr1993:
This was implemented in #355. Closing.

PR #228, opened 2019-05-08
Refactor checked and as_prover

mrmr1993:
This PR
* adds a `Types.Types` module type, to give a standard way to address types across the functored modules
* generalises `Provider.t`, moving `Provider.run` into `As_prover` so that it can work across different implementations
* moves `Snark0.Runner` into `Checked_runner`, adjusts the interface so that it forms a valid `Checked_intf.Basic`, which we will use (a derivative of) to create the imperative interface
* removes some unused definitions
* functors `Snark0.Make_basic` over `As_prover.S`
* moves `As_prover.Basic`/`S` into a dedicated `As_prover_intf` module

~~There is a small amount more work to be done to unpick the interdependencies between `Checked` and `As_prover`, and to integrate `Typ` into the functor. Once this is done, we will be able to generate the imperative API by just swapping out `Checked` with a replacement module in the functor.~~

Edit: this is now done, all that's left is to implement the interfaces (tweaking them where necessary)


src/as_prover.mli
```diff
@@ -29,6 +29,18 @@ module type Basic = sig
     -> 'var
     -> ('value, 'f field, 'prover_state) t
 
+  module Provider : sig
```
imeckler: Is there a reason you moved this into this module? Some kind of dependency or something?

mrmr1993: It used to depend on the specific `As_prover0.t` that we used as the canonical `As_prover.t`. To functor over the `As_prover` interface, it also needs to depend on the specific instance of `As_prover`, and so this seemed like a logical place to put it because of its dependency on the `As_prover` runner.

PR #233, opened 2019-05-10
[Meja] Add support for bare handlers

mrmr1993:
This PR
* adds parsing support for `handler _ => { _ }` in statements and in expressions
  - a handler at the end of an expression block is a parse error, since it de-sugars to a let
* generalises the request handler code and moves it into `Codegen`
* generalises the OCaml code generation for handlers
* adds typechecker support

Note: generating instances for these is currently disabled, pending row subtyping and PR #224.

### DIFF
--- a/meja/tests/handler.meja
+++ b/meja/tests/handler.meja
@@ -1,0 +1,14 @@
+request('a) A;
+
+handler A => {
+  unhandled;
+};
+
+request('a) B('a);
+
+let x = {
+  handler B(y) => {
+    y;
+  };
+  ();
+};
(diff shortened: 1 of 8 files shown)


meja/tests/handler.meja
```diff
@@ -0,0 +1,14 @@
+request('a) A;
+
+handler A => {
```
imeckler: Is the assumption that each request has exactly one handler? I would expect something more like

```reasonml
let handle_A = handler {
  | A => unhandled
};
```
which would compile to
```ocaml

let handle_A = function
  | With {request= A; respond} ->
      let unhandled = Snarky.Request.unhandled in
      unhandled
  | _ ->
      Snarky.Request.unhandled
```

though then you have to "register them" separately.

mrmr1993: Done. I'm going to merge this into #273 so I can get it integrated with the stdlib.

PR #234, opened 2019-05-10
[Meja] Extend Ast_build to support all types and type_decls

mrmr1993:
This PR cherry-picks some commits from #224, committing only the changes to `Ast_build`, and only when they don't depend on other changes.

Changes are:
* functions for the building remaining types that weren't previously supported
* functions for building type declarations

### DIFF
--- a/meja/src/ast_build.ml
+++ b/meja/src/ast_build.ml
@@ -2,7 +2,7 @@ open Core_kernel
 open Parsetypes
 
 module Loc = struct
-  let mk ?(loc = Location.none) x = Location.mkloc x loc
+  let mk ?(loc = Location.none) x : Location.t = Location.mkloc x loc
 
   let map x ~f = {Location.loc= x.Location.loc; txt= f x.Location.txt}
 end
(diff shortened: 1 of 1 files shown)


psteckler:
I think most of the Coda codebase uses `create` for building instances of module types, like `Foo.t`. Hee, `mk` is not building module types, but it's a similar task. Maybe it's worth following that here (or not).

It might be useful for the reader's sake to indicate what these functions are building, so that `Type_decl.mk` has the annotation `: Parsetypes.type_decl`.

mrmr1993:
For consistency, I've been trying to mirror the OCaml compiler's `Ast_builder` API format, since we use both APIs in similar ways for similar things in the codebase. Nothing from the Coda codebase should be calling into this, so I'd feel better keeping it close to the OCaml compiler format.

> It might be useful for the reader's sake to indicate what these functions are building, so that `Type_decl.mk` has the annotation `: Parsetypes.type_decl`.

I'll do this shortly, thanks.

PR #235, opened 2019-05-10
[Meja] Move common AST types out into their own module

mrmr1993:
This PR cherry-picks some commits from #224, committing only the changes to `Ast_types`/`Ast_print`, and only when they don't depend on other changes.

This PR is a no-op: it just separates out some code that the typechecker's types and the parsetree's types both depend on.


mrmr1993:
Closing, this was included in #224.

PR #236, opened 2019-05-10
[Meja] Move the initial environment into its own module

mrmr1993:
This PR
* moves `Initial_env` into its own module
* re-implements `Initial_env` using `Ast_build`

This makes the initial environment stable under changes to the types in `Parsetypes` by using the `Ast_build` wrapper.

### DIFF
--- a/meja/src/envi.ml
+++ b/meja/src/envi.ml
@@ -356,13 +356,13 @@ module Scope = struct
         Some (Immediate (f (find_module ~loc lid2 resolve_env scopes)))
 end
 
-let empty_resolve_env () =
+type t =
+  {scope_stack: Scope.t list; depth: int; resolve_env: Scope.t resolve_env}
+
+let empty_resolve_env : Scope.t resolve_env =
   { type_env= TypeEnvi.empty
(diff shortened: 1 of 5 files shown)


meja/src/envi.ml
```diff
@@ -356,7 +356,7 @@ module Scope = struct
         Some (Immediate (f (find_module ~loc lid2 resolve_env scopes)))
 end
 
-let empty_resolve_env =
+let empty_resolve_env () =
```
psteckler: Could this be left unthunked?

mrmr1993: Fixed.

meja/src/initial_env.ml
```diff
@@ -0,0 +1,118 @@
+(** The default initial environment. *)
+open Envi
+
+open TypeDecl
+open Ast_build
+
+(** The built-in types. These match the OCaml built-ins. *)
+module TypeDecls = struct
+  open Type
+  open Type_decl
```
psteckler: Slightly worrying there's a `TypeDecl` and a `Type_decl`.

PR #237, opened 2019-05-10
added a modify_or_get function to merkle tree

ghost-not-in-the-shell:
In this PR, I added a `modify_or_get function` to Merkle Tree. The implementation basically is the same as `modify`, but it would also return the old value back.

### DIFF
--- a/src/merkle_tree.ml
+++ b/src/merkle_tree.ml
@@ -456,6 +456,32 @@ struct
     in
     implied_root next_entry_hash addr0 prev_path
 
+  let%snarkydef_ modify_or_get_req ~(depth : int) root addr0 ~f :
+      (Hash.var * Elt.var, 's) Checked.t =
+    let open Let_syntax in
+    let%bind prev, prev_path =
+      request_witness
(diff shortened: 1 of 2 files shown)


src/merkle_tree.ml
```diff
@@ -456,6 +456,32 @@ struct
     in
     implied_root next_entry_hash addr0 prev_path
 
+  let%snarkydef_ modify_or_get_req ~(depth : int) root addr0 ~f :
```
mrmr1993: This and `modiify_req` look to be nearly identical.. can we combine them and just have `modify_req` discard the `prev` part of the return value? They also both seem to include the entire definition of `get_req`, which should probably be a function call instead.

ghost-not-in-the-shell: Sure, I agree those two functions look really similar. And `modify_or_get_req` besides doing the modification, also returns the old value. I delete the origin `modify_req` and replace it with the new one. I did it on a different branch though. https://github.com/o1-labs/snarky/pull/246. Please take a look at that one.

PR #240, opened 2019-05-14
[Meja] Don't carry a location with types in the environment

mrmr1993:
This PR
* removes the dependency on `type_loc` from the environment code
* tweaks the environment functions to accept a `loc` parameter for location
* inlines the `Tvar` case from `Envi.Types.import` into `Envi.Types.refresh_vars`, throwing an `AssertionError` if the passed type isn't a variable
  - this is the only logic change in this PR
  - this change lets us get rid of a cascade of otherwise-unnecessary location parameters
* associates all of the types with `Location.none` when they are created as part of the evironment

These changes make it simpler to split `type_expr` into a parsetree version and a typechecker/environment version, as in #224.

### DIFF
--- a/meja/src/codegen.ml
+++ b/meja/src/codegen.ml
@@ -111,7 +111,7 @@ let typ_of_decl env decl =
                 { decl with
                   tdec_desc=
                     TAlias
-                      (Envi.TypeDecl.mk_typ ~loc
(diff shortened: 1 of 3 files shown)


mrmr1993:
Closing, #224 supersedes this.

PR #241, opened 2019-05-14
[Meja] Rework type declaration imports

mrmr1993:
This PR
* reworks type declaration imports so that implicit arguments are calculated correctly
* stops associating the type declarations in the environment with a location
* ensures that all type expressions within a type declaration are imported properly from the parsetree into the environment


mrmr1993:
Closing, this logic was imported into the final version of #224, and the function will be stripped from `Envi` in a future PR.

PR #242, opened 2019-05-14
[Meja] Add some signature items

mrmr1993:
This PR adds signature items for the remaining statements that don't already have one.

This lets us store a signature for each module, which will be useful for debugging, error messages, and (eventually) the toplevel interpreter.


mrmr1993:
> Can you add an associated test?

Done!

PR #244, opened 2019-05-14
[Meja] Add a --compiler-backtraces flag

mrmr1993:
This PR
* adds a `--compiler-backtraces` flag, to give backtraces without recompiling Meja
* switches the flag on in the tests if the `MEJA_BACKTRACE` environment variable is present

### DIFF
--- a/meja/meja.ml
+++ b/meja/meja.ml
@@ -70,6 +70,7 @@ let main =
   let meji_files = ref [] in
   let cmi_files = ref [] in
   let cmi_dirs = ref [] in
+  let exn_backtraces = ref false in
   let arg_spec =
     [ ( "--ml"
       , Arg.String (set_and_clear_default ocaml_file)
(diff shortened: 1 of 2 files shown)


meja/meja.ml
```diff
@@ -126,6 +131,7 @@ let main =
           file := Some filename )
     usage_text ;
   let env = Initial_env.env in
+  Printexc.record_backtrace !exn_backtraces ;
```
psteckler: should recording happen even when `--compiler-backtraces` is not enabled?

mrmr1993: The `record_backtrace` takes a `bool` that sets recording on or off. `exn_backtraces` is `false` if the flag isn't passed, so it should only record with the flag

psteckler: oh, of course, thanks

PR #246, opened 2019-05-14
Tags for merkle tree

ghost-not-in-the-shell:
In this PR, I add a `Tag` to merkle_tree.ml. The tag would be used in the `Get_element` request to determine the target merkle_tree.

### DIFF
--- a/src/merkle_tree.ml
+++ b/src/merkle_tree.ml
@@ -149,7 +149,7 @@ let insert hash compress t0 mask0 address x =
 
 let ith_bit n i = (n lsr i) land 1 = 1
 
-let update ({hash; compress; tree= tree0; depth} as t) addr0 x =
+let update ({hash; compress; tree= tree0; depth; _} as t) addr0 x =
   let tree_hash = tree_hash ~default:(hash None) in
   let rec go_non_empty tree i =
     match tree with
@@ -408,17 +408,6 @@ struct
     let typ ~depth : (var, value) Typ.t = Typ.(list ~length:depth Hash.typ)
   end
 
-  module Tag = struct
-    type value = [`Curr_ledger | `Epoch_ledger]
-
-    type var = Boolean.var
-
-    let typ : (var, value) Typ.t =
-      Typ.transport Boolean.typ
-        ~there:(function `Curr_ledger -> false | `Epoch_ledger -> true)
-        ~back:(function false -> `Curr_ledger | true -> `Epoch_ledger)
-  end
-
   let implied_root entry_hash addr0 path0 =
     let rec go height acc addr path =
       let open Let_syntax in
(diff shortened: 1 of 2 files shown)


src/merkle_tree.ml
```diff
@@ -408,6 +408,17 @@ struct
     let typ ~depth : (var, value) Typ.t = Typ.(list ~length:depth Hash.typ)
   end
 
+  module Tag = struct
+    type value = [`Curr_ledger | `Epoch_ledger]
```
mrmr1993: I'm not sure that this needs to live in Snarky, it seems quite Coda-specific.

src/merkle_tree.ml
```diff
@@ -408,6 +408,17 @@ struct
...
+
+    type var = Boolean.var
+
+    let typ : (var, value) Typ.t =
```
mrmr1993: I think this would amount to the same using the [`Enumerable`](https://github.com/o1-labs/snarky/blob/master/src/enumerable.mli) interface, might be cleaner to do it that way.

src/merkle_tree.ml
```diff
@@ -426,19 +437,24 @@ struct
...
+      (Hash.var * Elt.var, 's) Checked.t =
     let open Let_syntax in
     let%bind prev, prev_path =
       request_witness
         Typ.(Elt.typ * Path.typ ~depth)
         As_prover.(
-          map (read (Address.typ ~depth) addr0) ~f:(fun a -> Get_element a))
+          Let_syntax.(
+            let%map addr = read (Address.typ ~depth) addr0
+            and tag = read Tag.typ is_chain_voting in
```
mrmr1993: This `read` would be better living outside the code for this and being more closely integrated with the handler. I would feel much more comfortable with something like:

```ocaml
let%bind tag =
  As_prover.(Ref.create (read Tag.typ is_chain_voting))
in
let code = (* code that calls modify_ref *) in
let%bind (* variables here *) =
  handle code As_prover.(fun (With {request; respond}) ->
    match request with
    | Get_element addr ->
        let%bind tag = Ref.get tag in
        (* handler code... *)
    | _ -> unhandled)
in
(* more code *)
```
at the callsite and removing this completely.

src/merkle_tree.mli
```diff
@@ -110,13 +120,13 @@ module Checked
...
-
   val modify_req :
-       depth:int
+       is_chain_voting:Boolean.var
+    -> depth:int
     -> Hash.var
     -> Address.var
     -> f:(Elt.var -> (Elt.var, 's) Checked.t)
-    -> (Hash.var, 's) Checked.t
+    -> (Hash.var * Elt.var, 's) Checked.t
```
mrmr1993: I'm not sure that overwriting this is the way we should go here, a new name (`fetch_and_update`?) and keeping an implementation of the original would cause less friction. Would be good to have a doc-comment clarifying the return value too.

src/merkle_tree.ml
ghost-not-in-the-shell: I agree with you. I will put the logic inside coda then.

ghost-not-in-the-shell: I am confused about this code snippet. I thought the second argument to `handle` should be a `Handler.t` . But `let%bind tag = Ref.get tag in` would be of type `As_prover.t`.

I am still learning the basics of snarky. Would you please explain what's happening here?

mrmr1993: Sorry, I got my wires a bit crossed on this. I think the code has to be

```ocaml
let tag = ref None in
let%bind tag =
  as_prover As_prover.(
    let%map is_chain_voting = read Tag.typ is_chain_voting in
    tag := Some chain_voting
  )
in
let code = (* code that calls modify_ref *) in
let%bind (* variables here *) =
  handle code As_prover.(fun (With {request; respond}) ->
    match request with
    | Get_element addr ->
        let tag = Option.value_exn !tag in
        (* handler code... *)
    | _ -> unhandled)
in
(* more code *)
```

I think there's probably value in us being able to pass `As_prover` blocks as handlers, so I'll make a PR for that tomorrow.

ghost-not-in-the-shell: This works. Thanks a lot.

PR #253, opened 2019-05-20
[Meja] Only resolve implicit arguments when the instance is a supertype

mrmr1993:
This PR
- adds an `is_subtype` method
- uses `is_subtype` to check whether instances are general enough to catch all the cases necessary
  * this fixes a bug (which made my computer OOM :sweat_smile:) that would recursively try and fill an implicit variable of type `'a` with an instance of type `poly('a)`.

### DIFF
--- a/meja/src/typechecker.ml
+++ b/meja/src/typechecker.ml
@@ -145,6 +145,16 @@ let check_type ~loc env typ constr_typ =
   | () ->
       ()
 
+(** [is_subtype ~loc env typ ~of_:ctyp] returns whether [typ] is a subtype of
+    [ctyp], instantiating any variables in [ctyp] to those they match with in
+    [typ].
+
+    If this function returns [false], the [ctyp] value *must not* be used,
+    since its variables may have been instantiated incorrectly. A type
+    containing only fresh variables should be used.
+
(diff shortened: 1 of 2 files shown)


meja/src/typechecker.ml
```diff
@@ -145,6 +145,101 @@ let check_type ~loc env typ constr_typ =
   | () ->
       ()
 
+(** [is_subtype ~loc env typ ~of_:ctyp] returns whether [typ] is a subtype of
+    [ctyp], instantiating any variables in [ctyp] to those they match with in
+    [typ].
+
+    If this function returns [false], the [ctyp] value *must not* be used,
```
bkase: a bit scary, would be nice if you could instead return an option or something, but I imagine the mutability here is pretty ingrained

mrmr1993: I was hitting some performance and consistency problems using a model without mutation. Currently I'm working towards making all the typechecking use mutability, to simplify the algorithms and give us speed and memory-usage improvements.

Once this is done (and the stdlib is working properly), I'm going to add backtracking, so we can recover a working environment when things like this happen. For now, I agree that this is worrying, but I don't have a better solution before the prereqs are in place..

Thanks for approving.

PR #255, opened 2019-05-20
[Meja] codegen typ generation

mrmr1993:
This PR
* switches `Codegen` to produce parsetree-like code instead of fake typed tree code
* adds some stub interface files to let the new `Codegen` code compile
* updates the expected output

This includes PRs #253 and #254.

Quirks:
* At present, we generate too many implicit variables when generating a `Typ.t`. In particular, the instances used for `alloc` and `check` depend only on the `var` and `value` parameters respectively, and the typechecker doesn't know that these variables can be unified.
  - I'm working on a re-work of the unification and depth logic, which will make these identifiable by their lower depth, as well as giving a nice speed-up.

### DIFF
--- a/meja/tests/records.ml
+++ b/meja/tests/records.ml
@@ -35,14 +35,10 @@ include struct
                     Typ.Alloc.return {a; b; c} ) ) )
     ; Typ.check=
         (fun {a; b; c; _} ->
-          (fun x f -> f x)
-            ((Typ.check __implicit1__) c)
-            (fun c ->
-              (fun x f -> f x)
-                ((Typ.check __implicit2__) b)
-                (fun b ->
-                  (fun x f -> f x) ((Typ.check __implicit3__) a) (fun a -> ())
-                  ) ) ) }
+          (Typ.check __implicit1__) c ;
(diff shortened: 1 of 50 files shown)


meja/tests/records.ml
```diff
@@ -40,30 +56,46 @@ module X = struct
   include struct
     type 'a t = {a: 'a; b: 'a; c: 'a}
 
-    let typ : (_, _ t) Typ.t =
+    let typ __implicit24__ __implicit23__ __implicit22__ __implicit21__
```
imeckler: Shouldn't this just have one implicit argument?

imeckler: Ah I guess this is fixed in https://github.com/o1-labs/snarky/pull/256 ?

mrmr1993: Not quite, #256 is part of the fix, but I'm not quite finished with the unification changes to make it work properly.

I've spun this out as its own PR because the old version was fragile and needed updating with nearly every typechecker change, whereas the new version generates a parsetree and lets the typechecker handle it as normal.

PR #258, opened 2019-05-22
[Meja] Add row types

mrmr1993:
This PR adds row types to the parser and typechecker.

Currently, the row types cannot be instantiated, since they are incompatible with OCaml's row types. Instead, these will be a phantom type parameter (not exported to OCaml), which classifies which constructors of an open type (notably either `request` or `exn`) a value may represent, and allow other types to express relations between them in terms of these phantom types.

For example, request handlers will be of the type `(_, _, ['a]) Checked -> (_, _, [['a] - [Handled_request]]) Checked)`.

### DIFF
--- a/meja.opam
+++ b/meja.opam
@@ -14,6 +14,7 @@ depends: [
   "ocaml-compiler-libs"
   "ppxlib"
   "ppx_jane"
+  "ppx_deriving"
   "dune"                {build & >= "1.0+beta12"}
 ]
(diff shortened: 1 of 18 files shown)


imeckler:
The intended use for these is for tracking effects right? Maybe it would be better to have a separate "kind" for requests to disallow stuff that doesn't make sense in OCaml? Or is that worse ergonomically? Maybe we should write some "real" programs and see what looks good/makes sense.

PR #262, opened 2019-05-30
Add a function to check that the keys in a keypair are compatible

mrmr1993:
This PR
* adds the `check`/`check_keys` functions to the snarky interface
* implements a fast check for the GG and BG backends, checking for equality between the shared component of the proving and verification keys (`delta_g2`)
* stubs the check for the GM backends to always return true
  - it would be good to find a cheap test for these as well


src/libsnark.ml
```diff
@@ -1481,6 +1485,17 @@ module Make_proof_system_keys (M : Proof_system_inputs_intf) = struct
...
         let t = stub sys in
         Caml.Gc.finalise delete t ; t
+
+    let of_keys =
+      let stub =
+        foreign (func_name "of_keys")
+          (Proving_key.typ @-> Verification_key.typ @-> returning typ)
+      in
+      fun pk vk ->
+        let t = stub pk vk in
```
cmr: I think we should `assert (check t)`.

src/camlsnark_c/libsnark-caml/libsnark/caml/caml_bn128.cpp
```diff
@@ -720,6 +720,18 @@ r1cs_gg_ppzksnark_keypair<ppT>* camlsnark_bn128_keypair_create(
   return new r1cs_gg_ppzksnark_keypair<ppT>(res);
 }
 
+r1cs_gg_ppzksnark_keypair<ppT>* camlsnark_bn128_keypair_of_keys(
+    r1cs_gg_ppzksnark_proving_key<ppT>* pk,
+    r1cs_gg_ppzksnark_verification_key<ppT>* vk) {
+  r1cs_gg_ppzksnark_proving_key<ppT> pk_(*pk);
```
cmr: The proving key is quite large. It is unfortunate that this copies it.

mrmr1993:
Stale, we don't use libsnark any more

PR #263, opened 2019-05-30
Remove typo in pattern variable printing

mrmr1993:
This makes pattern variables print properly, e.g. `let {foo= bar} = x;` instead of `let {foo= 'bar} = x;`.


mrmr1993:
Closing. This is included in #264.

PR #268, opened 2019-06-06
[Meja] Fix unification for GADTs

mrmr1993:
This PR
* (parsing changes for the tests:)
  * fixes the parsing of infix operators in `switch` statements
  * relaxes the restriction on expression type annotations, so that any expression can be annotated
    - in particular, this now means that we can annotate constructors, functions, etc.
  * changes the presence of type annotation expressions, so that they can be used everywhere a normal bracketed expression would be, as expected
* adds the concept of a generic depth, which unification doesn't change
* ensures that newtypes are fixed at dynamic depth
* coerces newtypes to type variables within `switch` pattern matches, so that they can be destructed by GADTs
  - this uses the generic depth to identify whether a GADT has eliminated them or not: GADTs lower the levels of their type parameters, so we can detect at the pattern level whether a GADT specialised the newtype
* adds some tests

These will be used in type parameters for request handlers going forward, so that we can handle requests containing GADTs.

### DIFF
--- a/meja/src/type0.ml
+++ b/meja/src/type0.ml
@@ -33,7 +33,8 @@ and type_decl =
   ; tdec_params: type_expr list
   ; tdec_implicit_params: type_expr list
   ; tdec_desc: type_decl_desc
-  ; tdec_id: int }
+  ; tdec_id: int
+  ; tdec_is_newtype: bool }
 
 and type_decl_desc =
   | TAbstract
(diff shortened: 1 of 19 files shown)


meja/src/type0.ml
```diff
@@ -1,6 +1,8 @@
 open Core_kernel
 open Ast_types
 
+let generic_depth = 1000000
```
imeckler: We could use a variant here if it's not too annoying ? Like

```ocaml
module Depth : sig
  type t [@@deriving eq]
  val generic : t
  val zero : t

  val succ :t -> t 
end  = struct
  type t = Generic | Concrete of int
  let zero = Concrete 0
  let generic = Generic

  let succ = function Generic -> Generic | Concrete n -> Concrete (n + 1)
end

PR #287, opened 2019-06-25
[Meja] Add support for applying optional arguments with labels directly

mrmr1993:
This PR
* tweaks unification so that `f(~x= ..., ...)` typechecks when we have `f : ?x:... -> ...`
* adds some tests


meja/src/initial_env.ml
```diff
@@ -115,3 +115,5 @@ module Type = struct
 
   let list a = TypeDecl.mk_typ list ~params:[a] env
 end
+
+let () = Envi.Type.mk_option := Type.option
```
imeckler: :)

PR #289, opened 2019-06-25
[Meja] Fix parsing bug so that request and handler may be function argument labels

mrmr1993:


imeckler:
Hm -- do you know if there's any reason why in OCaml keywords can't be labels? Seems like an unnecessary restriction no?

mrmr1993:
>  do you know if there's any reason why in OCaml keywords can't be labels? Seems like an unnecessary restriction no?

Can't think of one, I'll try and make a PR over there at some point :smile:

mrmr1993:
On reflection, the reason that this isn't done in OCaml is that there is no clear distinction between function application `f a` and expressions involving keywords `let x`. Especially within the toplevel, this would lead to parser ambiguity.

More importantly, we already have this overloading everywhere except labels. This PR is really a typo fix in the parser (`LIDENT` -> `lident`).

We shouldn't need to worry about `handler` and `request` being identifiers, because
* function application is `f(a)`, and the next token must be separated by some non-whitespace character
* `request` (at the expression level) takes the form `request(type) Constructor`
  - technically `()` is a constructor, so there is ambiguity in `request(type) ()`. This is a bug separate to this PR; I will disallow `()`, `true`, `false`, etc. as request names later, since I don't think we ever want them anyway.
  - there is no other overlap between expressions involving identifiers and uses of `request`.
* `handler` is only used in the form `with handler` currently, so it is always unambiguous
  - the extra syntax (which I will re-add post the reworked stdlib) is also always unambiguous:
```
handle {
  | pattern => expression
  ...
  }
```

PR #292, opened 2019-06-25
[Meja] Add support for if (...) { ... } else { ... } syntax

mrmr1993:
This PR adds support for `if` statements. In particular, this supports the forms:
```js
/* if with no else statement, return value must be unit */
if (...) {
  ...
}

/* classic if-else */
if (...) {
  ...
} else {
  ...
}

/* combined if-else-if (can terminate in an `else if`, in which case it must have type `unit`) */
if (...) {
  ...
} else if (...) {
  ...
} else if (...) {
  ...
} ...
```


imeckler:
Can you do `let x = if (b) { 1; } else if (b') { 2; } else { 3; }` ? For some reason I remember trying this and it didn't work -- not a big deal, just wondering

PR #306, opened 2019-07-01
Add 4.08 support to snarky

mrmr1993:
This PR
* updates the `*.opam` files to reflect the actual version constraints
* changes the code generated by `snarky_ppx` to produce `Stdlib` instead of `Pervasives` (which is deprecated in 4.08)

There are still compilation errors in meja with 4.08, but they're slightly more involved because it interfaces with `ocaml-compiler-libs`. I haven't found a good way to support 4.07 and 4.08 simultaneously yet, so (for now) those migrations will have to happen at the time of the 4.07->4.08 change..

### DIFF
--- a/interval_union.opam
+++ b/interval_union.opam
@@ -10,13 +10,13 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "core_kernel"
+  "core_kernel" {>= "v0.12" & < "v0.13" }
   "ppx_deriving"
   "ppx_jane"
   "bisect_ppx"
-  "jbuilder"                {build & >= "1.0+beta12"}
+  "dune"                {build & >= "1.6"}
 ]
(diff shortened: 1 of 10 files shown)


interval_union.opam
```diff
@@ -10,13 +10,13 @@ build: [
...
-  "core_kernel"
+  "core_kernel" {>= "v0.12" & < "v0.13" }
   "ppx_deriving"
   "ppx_jane"
   "bisect_ppx"
-  "jbuilder"                {build & >= "1.0+beta12"}
+  "dune"                {build & >= "1.6"}
 ]
-available: [ ocaml-version >= "4.04.1" ]
+available: [ ocaml-version >= "4.07.0" ]
```
psteckler: in some `.opam` files, the version is `4.07.1`, here it's `4.07.0` ... should these be uniform?

PR #308, opened 2019-07-03
Functor the core FFI bindings over Ctypes.FOREIGN

mrmr1993:
This PR functors the commonly used FFI bindings over `Ctypes.FOREIGN`, so that we can generate C stubs using the Ctypes API and remove the majority of our libffi overhead.

This doesn't do the actual generation yet, that will be in a follow-up PR.

### DIFF
--- a/src/backend_types.mli
+++ b/src/backend_types.mli
@@ -6,17 +6,26 @@ module type Prefix_intf = sig
   val prefix : string
 end
 
+module type Foreign_types = sig
+  type 'a return
+
+  type 'a result
+end
+
 module type S = sig
+  include Foreign_types
+
   type t
 
   val typ : t typ
 
   val func_name : string -> string
 
-  val delete : t -> unit
+  val delete : (t -> unit return) result
 end
 
-module Make_foreign (M : Prefix_intf) : S
+module Make_foreign (F : Ctypes.FOREIGN) (M : Prefix_intf) :
+  S with type 'a return := 'a F.return and type 'a result := 'a F.result
 
 module type Field_prefix_intf = sig
   include Prefix_intf
@@ -27,7 +36,11 @@ end
 module type Field_constrained = sig
   type 'field t
 
-  module Make (M : Field_prefix_intf) : S with type t = M.field t
+  module Make (F : Ctypes.FOREIGN) (M : Field_prefix_intf) :
+    S
+    with type t = M.field t
+     and type 'a return := 'a F.return
+     and type 'a result := 'a F.result
 end
 
 module Var : Field_constrained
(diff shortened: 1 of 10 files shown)


src/backend_types.mli
```diff
@@ -27,7 +36,11 @@ end
 module type Field_constrained = sig
   type 'field t
 
-  module Make (M : Field_prefix_intf) : S with type t = M.field t
+  module Make (F : Ctypes.FOREIGN) (M : Field_prefix_intf) :
+    S
+    with type t = M.field t
```
imeckler: maybe have F be a submodule inside of S so you can := both types at once with a module := ?

mrmr1993: @imeckler nice, I'll do this, thanks. Would be nice to do that as a separate PR though, I'd like to get this in so I can build get rid of the libffi stuff on Friday during the holiday!

src/long_vector.ml
```diff
@@ -1,11 +1,19 @@
 open Ctypes
 
-include Vector.Make (struct
-  let prefix = "camlsnark_long_vector"
+module Bindings =
```
imeckler: Should we compress abstract these two functor calls into one?

mrmr1993:
> Is there some way we can make this less verbose?

I really intended this PR to be a sanity check that we still have the same bindings implemented in the same way, but separating the C bindings from the OCaml wrappers. I should probably have stated as much in the description, sorry.

Concretely, the plan is:
* split the code in libsnark into the C bindings (functored over `Ctypes.FOREIGN`) and the OCaml layer above them (functored over the bindings) [this PR].
  - everything should function exactly the same, checking that this is true is the main point of breaking this out as its own PR
* move the bindings into the `camlsnark_c` library, use `ctypes_cstubs` to generate direct C bindings in OCaml there, and use those non-libffi bindings to instantiate the OCaml parts left in `libsnark.ml`
  - this will move all of the `foreign` calls, binding signatures, etc. out of `libsnark.ml`, reducing the line count substantially
  - this will also be quite a large PR because we will also need to add headers for our current C bindings to link to the FFI, so I wanted it to be separate from this
* generate ctypes bindings for the actual libsnark types, switch over to using ctypes-generated allocations for the ffi instead of C++ `new`/`delete`
  - this should also remove the bulk of this PR's remaining code in `libsnark.ml`, which is dedicated to integrating the OCaml GC with these `new`/`delete` calls
  - quite a lot of the C code will need tweaking to work with the OCaml GC/not use `new`/`delete`

> It would suck for libsnark.ml (which is already super long) to gain 800 lines.

I had left the bindings in `libsnark.ml` for this PR because I thought it would be easier to check correctness while it's all still in one place. I'm very happy to move them to `camlsnark_c` now to reduce the line count, if you would prefer?

PR #315, opened 2019-07-12
Truthy values

mrmr1993:
This PR
* adds `Truthy` values type (0 - false, anything else - true) and the natural cheap operations that may be done with them
* adds a zero-constraint `not_equal` function for field elements
* adds a zero-constraint `not_equal` function to `Bitstring_checked`
* adds `Bitstring_checked.Assert.either_equal`
  - this costs the same as `Bitstring_checked.Assert.either`, but it lets us make conditional assertions
  - this also doesn't go through `Boolean.t`, so we don't have to compute the inverses of any field elements

The biggest motivator for this is the use of `Schnorr.Checked.verifies` in Coda, where
* we use 2 constraints for each chunk to create its `Boolean.t`
* we calculate the field inverse for each chunk to satisfy the constraints for each `Boolean.t`
* we use a further 2 constraints for the return value
* we use a further constraint and compute another field inverse for the final assertion in `Transaction_snark.check_signature`

By contrast, with this we could stop computing any inverses (which showed this up as a surprising hotspot during profiling) and more than halve the number of constraints for the assertion along the way.

Concretely, the changed code might look like:
```
    let%snarkydef assert_either_verifies ~or_ shifted ((_, h) as signature) pk m =
      verification_hash shifted signature pk m
      >>= Curve.Scalar.Checked.Assert.either_equal ~or_ h
```
```
    module Assert = struct
      ...
      let either_equal : or_:Boolean.t -> var -> var -> (unit, _) Checked.t =
       fun ~or a b ->
        Bitstring_checked.Assert.either_equal
          ~or_not:(Truthy.of_boolean (Boolean.not or_))
          (Bitstring.Lsb_first.to_list a)
          (Bitstring.Lsb_first.to_list b)
    end
```


imeckler:
Cool idea! Surely the inversions performed in curve operations are where the bulk of the inversions are happening though no?

mrmr1993:
> Surely the inversions performed in curve operations are where the bulk of the inversions are happening though no?

This is true, but the `equal` call was turning out much heavier than I had expected, and it seemed like something we could fix. It could also be that there's some inlining happening there that's confusing the profiler.

I'll run some tests to see if we see much difference.

mrmr1993:
Profiled runs without and with.

#### Using `/usr/bin/time`

Without:
```
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 > /dev/null
2152.94user 7.87system 12:49.13elapsed 281%CPU (0avgtext+0avgdata 911924maxresident)k
272inputs+80outputs (0major+3409644minor)pagefaults 0swaps

$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
122.58user 1.18system 2:03.86elapsed 99%CPU (0avgtext+0avgdata 154024maxresident)k
0inputs+80outputs (0major+655864minor)pagefaults 0swaps
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
123.07user 1.19system 2:04.34elapsed 99%CPU (0avgtext+0avgdata 153844maxresident)k
0inputs+80outputs (0major+655443minor)pagefaults 0swaps
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
122.10user 1.11system 2:03.35elapsed 99%CPU (0avgtext+0avgdata 153628maxresident)k
72inputs+80outputs (0major+655257minor)pagefaults 0swaps
```

With:
```
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 > /dev/null
2151.70user 8.03system 12:51.50elapsed 279%CPU (0avgtext+0avgdata 905028maxresident)k
536inputs+80outputs (0major+3463576minor)pagefaults 0swaps

$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
121.06user 1.21system 2:02.30elapsed 99%CPU (0avgtext+0avgdata 153816maxresident)k
8inputs+80outputs (0major+652186minor)pagefaults 0swaps
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
121.48user 1.18system 2:02.71elapsed 99%CPU (0avgtext+0avgdata 153660maxresident)k
8inputs+80outputs (0major+654303minor)pagefaults 0swaps
$ `which time` _build/default/app/cli/src/coda.exe tra -repeat 10 -witness-only > /dev/null
121.80user 1.35system 2:03.22elapsed 99%CPU (0avgtext+0avgdata 153972maxresident)k
8inputs+80outputs (0major+654796minor)pagefaults 0swaps
```

------

#### QAP stats

without:
```
* QAP number of variables: 240033
* QAP pre degree: 240554

* QAP number of variables: 154208
* QAP pre degree: 151344

* QAP number of variables: 13810
* QAP pre degree: 13806
```

with:
```
* QAP number of variables: 240027
* QAP pre degree: 240550

* QAP number of variables: 154208
* QAP pre degree: 151344

* QAP number of variables: 13810
* QAP pre degree: 13806
```

------

#### Per-run timing:

without:
```
$ dune exec coda -- tra -repeat 10 | grep "Total time was:"
[1] Total time was: 47.049670219421387s
[2] Total time was: 46.393154859542847s
[3] Total time was: 46.62542462348938s
[4] Total time was: 46.3239483833313s
[5] Total time was: 46.557117223739624s
[6] Total time was: 47.474007129669189s
[7] Total time was: 46.690899133682251s
[8] Total time was: 46.388137102127075s
[9] Total time was: 46.3362762928009s
[10] Total time was: 46.437490701675415s
```
Average: `46.62761256694794`
Average of median 4 values (via haskell):
```haskell
Prelude Data.List> (/4) . sum . take 4 . drop 3 . sort $ vals_without
46.503296852111816
```

with:
```
$ dune exec coda -- tra -repeat 10 | grep "Total time was:"
[1] Total time was: 46.668546676635742s
[2] Total time was: 46.411561250686646s
[3] Total time was: 46.0884211063385s
[4] Total time was: 46.515727043151855s
[5] Total time was: 46.506227493286133s
[6] Total time was: 46.129955530166626s
[7] Total time was: 46.356711387634277s
[8] Total time was: 46.184632778167725s
[9] Total time was: 46.343616008758545s
[10] Total time was: 46.491071224212646s
```
Average: `46.36964704990387`
Average of median 4 values (via haskell):
```haskell
Prelude Data.List> (/4) . sum . take 4 . drop 3 . sort $ vals_with
46.40073996782303
```

-----

#### Summary

* Makes ~1s difference for both witness generation and proving over 10 runs. That's about 100ms (~1%) speed-up in witness generation time, ~0.2% speed-up when we include proving time.
* Saves 6 variables, 4 constraints

PR #316, opened 2019-07-15
Validate curve points

imeckler:
This PR validates curve points when they are deserialized. We both check that the curve point is on the curve, and that it is on the prime order subgroup. 

In the case of G1, since the curve is itself prime order, the subgroup check is not required. 

In the case of G2, a curve point P is equal to `g + h` where `g` is in the order `p` subgroup and `h` is in its orthogonal complement (i.e., we factor G2 = G + H where G is the order `p` subgroup and h is in H). We can check that `P` is in the prime subgroup by computing `p * P = p g + p h = p h`. Since the order of H is not divisible by p, `p h` is the identity iff `h` is the identity iff `P` is in G.


src/libsnark.ml
```diff
@@ -412,8 +421,28 @@ struct
     let to_repr t : Repr.t =
       match to_affine t with None -> Zero | Some t -> Non_zero t
 
+    let subgroup_check t =
+      match Checks.subgroup with
+      | `No_check_required ->
+          ()
```
enolan: do you not need to check the point isn't the identity?

src/libsnark.ml
```diff
@@ -2701,19 +2930,59 @@ struct
...
                    (struct
                      let prefix = with_prefix Mnt4_0.prefix "g2"
                    end))
-                   (Mnt6_0.Field.Vector)
+                   (Fqe)
+                (struct
+                  let subgroup = `Check_subgroup_with_order Mnt4_0.field_size
+
+                  let on_curve (x, y) =
+                    let open Fqe in
```
mrmr1993: Might be worth using mutability here, as below?

src/libsnark.ml
```diff
@@ -2823,19 +3095,60 @@ struct
...
                      let prefix = with_prefix Mnt6_0.prefix "g2"
                    end))
-                   (Mnt4_0.Field.Vector)
+                   (Fqe)
+                (* TODO: Unify this with the Mnt4 code *)
+                (struct
+                  let subgroup = `Check_subgroup_with_order Mnt6_0.field_size
+
+                  let on_curve (x, y) =
+                    let open Fqe in
```
mrmr1993: Ditto

src/libsnark.ml
imeckler: See line 440. The identity is represented by a separate variant and can't be represented using affine coordinates.

src/libsnark.ml
imeckler: It requires adding mutable operators for the field extension multiplication and addition to libff. Addition is easy to do but multiplication is complicated because the algorithm requires additional storage space. Might still be possible to save some allocations though.

mrmr1993: Okay. If we notice it being a problem later, we can come back and do this.

PR #317, opened 2019-07-16
Add Typ.t's for non-snarkable values

mrmr1993:
This will let us avoid the hacks that we used in Meja during the
conference to handle string passing etc. between the worlds.

This slightly overloads the behaviour of `As_prover.Ref.t`, but for Meja's
purposes this will be invisible to the user anyway. Perhaps more
importantly, this type can be special-cased to give good error messages
when the user tries to use prover-generated data outside of a prover
block.

Generating code that uses `snarkable` will be slightly more awkward,
since we need to track when the same variable is read and then
re-stored, but it should be possible to make it work.  The important case
that this supports is code like:

```reason
...
let x = {data: 1000f, aux_data: 50};
let x = Prover { {x.. data= x.data * 10f} };
...
```

where the int `aux_data` might be used later in checked code as e.g. the
length of a loop

### DIFF
--- a/src/snark_intf.ml
+++ b/src/snark_intf.ml
@@ -290,27 +290,6 @@ module type Basic = sig
 
     val field : (Field.Var.t, field) t
 
-    val snarkless : 'a -> ('a, 'a) t
-    (** A do-nothing [Typ.t] that returns the input value for all modes. This
-        may be used to convert objects from the [Checked] world into and
-        through [As_prover] blocks.
-
-        This is the dual of [ref], which allows [OCaml] values from [As_prover]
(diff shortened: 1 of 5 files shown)


src/snark_intf.ml
```diff
@@ -289,6 +290,27 @@ module type Basic = sig
 
     val field : (Field.Var.t, field) t
 
+    val snarkless : 'a -> ('a, 'a) t
```
imeckler: Should we maybe put these into an "internal" module or something?

mrmr1993: Done, and added a comment to recommend avoiding them.

PR #322, opened 2019-07-17
Add generate_witness

mrmr1993:
This PR adds a `generate_witness` function to both the old- and new-style proving APIs. The field vectors that it returns can then be passed to alternative proving implementations.


src/snark_intf.ml
```diff
@@ -1058,6 +1058,17 @@ let multiply3 (x : Field.Var.t) (y : Field.Var.t) (z : Field.Var.t)
        [verification_key] overrides the argument given to {!val:create}, if
        any.
     *)
+
+    val generate_witness :
```
imeckler: Maybe a slightly different name since the "witness" is really the auxiliary input. I have no idea what it should be though :)

mrmr1993: I haven't been able to come up with a better name yet. I've tweaked the comment to clarify that we're only generating the `auxiliary_input`, do you think that makes it clear enough?

I've also tweaked the return type to make it a) harder to confuse the public and auxiliary inputs and b) to hopefully make it clearer that it's not returning just the auxiliary input in some strange form.

PR #325, opened 2019-07-18
backend functions for external prover

imeckler:
These are all the changes I needed when implementing the external prover wrapper

### DIFF
--- a/src/camlsnark_c/libsnark-caml/libsnark/caml/caml_curve.cpp.template
+++ b/src/camlsnark_c/libsnark-caml/libsnark/caml/caml_curve.cpp.template
@@ -649,6 +649,11 @@ void CURVE_PREFIX(r1cs_constraint_system_iter)(
   }
 }
 
+int CURVE_PREFIX(r1cs_constraint_system_num_constraints)(
+    r1cs_constraint_system<FieldT>* sys) {
+  return sys->constraints.size();
+}
+
 std::vector<FieldT>* CURVE_PREFIX(field_vector_create)() {
   return new std::vector<FieldT>();
 }
@@ -669,6 +674,14 @@ FieldT* CURVE_PREFIX(field_vector_get)(std::vector<FieldT>* v, int i) {
 
 void CURVE_PREFIX(field_vector_delete)(std::vector<FieldT>* v) {
   delete v;
+}
+
+void CURVE_PREFIX(field_vector_iter)(
+    std::vector<FieldT>* v,
+    void (*f)(const r1cs_constraint<FieldT> *)) {
+  for (std::vector<FieldT>::const_iterator i = v.cbegin(); i != v.cend(); i++) {
+    f(&*i);
+  }
 }
 
 // Begin ppzksnark specific code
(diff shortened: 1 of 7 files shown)


src/camlsnark_c/libsnark-caml/libsnark/caml/caml_curve.cpp.template
```diff
@@ -477,6 +481,40 @@ const linear_combination<FieldT>* CURVE_PREFIX(r1cs_constraint_c)(const r1cs_con
...
+        ca->at(i + sys->num_constraints()) = (i > 0 ? full_variable_assignment->at(i-1) : FieldT::one());
+    }
+    for (size_t i = 0; i < sys->num_constraints(); ++i)
+    {
+        ca->at(i) += sys->constraints[i].a.evaluate(*full_variable_assignment);
+        cb->at(i) += sys->constraints[i].b.evaluate(*full_variable_assignment);
+    }
+    for (size_t i = 0; i < sys->num_constraints(); ++i)
+    {
+        cc->at(i) += sys->constraints[i].c.evaluate(*full_variable_assignment);
```
mrmr1993: Can we lift this into the loop above?

src/camlsnark_c/libsnark-caml/libsnark/caml/caml_curve.cpp.template
```diff
@@ -477,6 +481,40 @@ const linear_combination<FieldT>* CURVE_PREFIX(r1cs_constraint_c)(const r1cs_con
...
+    const size_t degree)
+{
+    const size_t d_plus_1 = degree + 1;
+    std::vector<FieldT>* ca = new std::vector<FieldT>(d_plus_1, FieldT::zero());
+    std::vector<FieldT>* cb = new std::vector<FieldT>(d_plus_1, FieldT::zero());
+    std::vector<FieldT>* cc = new std::vector<FieldT>(d_plus_1, FieldT::zero());
+
+    for (size_t i = 0; i <= sys->num_inputs(); ++i)
+    {
+        ca->at(i + sys->num_constraints()) = (i > 0 ? full_variable_assignment->at(i-1) : FieldT::one());
```
mrmr1993: This seems like an unusual place to keep the variable assignments.. Is there a reason that we can't use the `full_variable_assignment` vector directly (and thus also remove the zero-padding at the end of `cb` and `cc`)?

src/camlsnark_c/libsnark-caml/libsnark/caml/caml_curve.cpp.template
```diff
@@ -611,6 +649,11 @@ void CURVE_PREFIX(r1cs_constraint_system_iter)(
   }
 }
 
+int CURVE_PREFIX(r1cs_constraint_system_num_constraints)(
+    r1cs_constraint_system<FieldT>* sys) {
+  return sys->constraints.size();
```
mrmr1993: Might be more consistent to use `sys->num_constraints()` here.

src/libsnark.ml
```diff
@@ -1138,6 +1166,21 @@ struct
...
+      let b = field evals_struct "b" Field_vector.typ in
+      let c = field evals_struct "c" Field_vector.typ in
+      seal evals_struct ;
+      let stub =
+        foreign
+          (func_name "evaluations")
+          (typ @-> Field_vector.typ @-> size_t @-> returning evals_struct)
+      in
+      fun t ~full_assignment ~degree ->
+        let res = stub t full_assignment (Unsigned.Size_t.of_int degree) in
```
mrmr1993: Should we assert here that `degree >= num_inputs + num_constraints + 1`? I'm concerned that this may be a source for segfaults/memory corruption later otherwise.

PR #328, opened 2019-07-19
First-class support for arrays

mrmr1993:
This PR
* adds a `Boolean.Array` module to `Snark_intf.*`
* adds a `Monad_sequence.Array` functor
* removes some unnecessary `Boolean.` prefixes for types in the `Boolean` module

This should make it much more straightforward to use `array`s, as well as making the matrix part of the tutorial much easier (although you would still need to construct `Monad_sequence.Array` for `Typ_monad.Store`/`Read`/`Alloc` manually to create the `Typ.t`).

### DIFF
--- a/src/monad_sequence.ml
+++ b/src/monad_sequence.ml
@@ -190,6 +190,14 @@ module Array
 
   let all_unit = iter ~f:(fun x -> x)
 
+  (* The following functions need to evaluate on every element, since
+     * we can't shortcut in the snark -- every constraint is either always
+       present or always absent -- so we need to evaluate [f] on every element
+     * [Bool.any]/[Bool.all] operate by calculating the sum of the array,
+       which requires an R1CS variable representing each element to be
+       available.
+  *)
+
   let existsi t ~f = mapi t ~f >>= Bool.any
 
   let exists t ~f = map t ~f >>= Bool.any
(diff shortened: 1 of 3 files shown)


src/monad_sequence.ml
```diff
@@ -125,3 +125,76 @@ module List
...
+          res := acc ;
+          y )
+    in
+    (!res, t)
+
+  let all = map ~f:(fun x -> x)
+
+  let all_unit = iter ~f:(fun x -> x)
+
+  let existsi t ~f = mapi t ~f >>= Bool.any
```
imeckler: Could you add a comment here (and to the list one) explaining that these are implemented using `map` instead of fold because `Bool.all/any` need the whole sequence to be optimally efficient in terms of constraints?

src/snark0.ml
```diff
@@ -460,6 +460,46 @@ struct
             in
             res
 
+      module Array = struct
+        let num_true (bs : var array) =
+          Array.fold bs ~init:(Cvar.constant Field.zero) ~f:(fun x y ->
+              Cvar.add x (y :> Cvar.t) )
+
+        let any = function
```
imeckler: Maybe it's too much abstraction but we could implement both of these with one higher-order function :)

src/monad_sequence.ml
mrmr1993: Done.

PR #335, opened 2019-07-26
[Meja] Prover blocks syntax

mrmr1993:
This PR adds the syntax for prover blocks.

As well as the expression form from the conference, this also adds a structure and signature form, so toplevel names, type declarations, etc. can be defined to be prover-only.

Edit: This doesn't actually *do* anything yet, it needs #330, #334, and a follow-up to both that integrates `Path.t` with modes before we can start segmenting the namespace using this.


mrmr1993:
Closing in favour of #344.

PR #337, opened 2019-07-26
Add a Field_unpacked module

### LINKED ISSUE #332: Unify "unpacked" and "packed" variables
Moved from https://github.com/CodaProtocol/coda/issues/121.

mrmr1993:
This PR adds the `Field_unpacked` module.

The idea is that
* the underlying representation is just a bitstring in the R1CS
* it is always 'free' in the R1CS to get from a `Field_unpacked.t` to a `Field.t` **and** to convert to a `Bitstring_checked.t`/`Boolean.t list`
  - most importantly, it makes it easier to avoid unpacking the same field element twice
* we avoid the nuances of the `Number` implementation, where multiple calls to `to_bits` are free in some cases and expensive in others, depending on what operations have been done to it

This fixes #332.

### DIFF
--- a/src/snark_intf.ml
+++ b/src/snark_intf.ml
@@ -714,8 +714,8 @@ let multiply3 (x : Field.Var.t) (y : Field.Var.t) (z : Field.Var.t)
 
       val unpack_full :
         Var.t -> (Boolean.var Bitstring_lib.Bitstring.Lsb_first.t, _) Checked.t
-      (** [unpack x ~length] returns a list of R1CS variables containing the
-          bits of [x].
+      (** [unpack_full x] returns a list of R1CS variables containing the bits
+          of [x], with the least significant bit of x first.
       *)
 
       val choose_preimage_var :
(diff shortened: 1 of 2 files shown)


mrmr1993:
> I think it would be better if we converted back to bits lazily (and used mutation) because there are sometimes where, e.g., you do an `if_` but don't actually need to go back to bits. Let me know what you think.

I'd quite like to keep this as is, IMO it makes sense to have a way to track whether we have 'paid' to unpack the field element, and the type system seems like a good place to do that. I would be quite uncomfortable if the offering in the standard library was a function that sometimes added hundreds of constraints and sometimes added none, depending on some hidden state.

Unless you feel strongly against it, I'd prefer to implement another type on top of this that does the caching. This way, users that care about tracking whether they've paid the constraints or not can access the information (but have to do e.g. `Field.Checked.if_ ~then_:(Unpacked_field.pack x) ~else_:(Unpacked_field.pack y)` to avoid paying another unpacking cost), and users that don't can have the benefits of the automatic caching.

mrmr1993:
We're no longer focused on R1CS, so this makes less sense. Stale, closing.

PR #343, opened 2019-07-30
[Meja] Use modes in Ident.Table.t

mrmr1993:
This PR
* integrates mode filtering with `Ident.Table.t`
* uses the mode filtering to fetch names according to which mode the current scope is for


mrmr1993:
Closing in favour of #344.

PR #344, opened 2019-07-30
[Meja] Prover mode

mrmr1993:
This PR implements prover mode, using `Prover { ... }` syntax to delimit prover blocks in expressions and prover-only type definitions etc. at the statement/signature level.

This replaces #335 and #343, and includes the final few changes to integrate the prover block syntax with the typechecker.


mrmr1993:
> could add an example of using the `Prover` block in such a way that it's translated to an exists

This PR is just the scoping changes; I'll follow up with the `exists`/`as_prover` generation, but it's going to live separately in the Meja->OCaml converter, so that
* we can do something different when we add a rust backend
* the meja AST is still valid after typechecking (which wasn't true in our conference version)
* we can do impromptu `Typ.t` generation in the separate conversion pass, so we avoid the pain of maximum tuple size, having hacks to pass strings into prover scopes, etc.

PR #346, opened 2019-08-02
Tweak the dummy verification key function

imeckler:
This PR tweaks the dummy verification  key backend function to have any `Fqk` elements be "all ones" so to speak. The reason is the following:

When hashing a verification key, we want to compress it down to as much data as possible.
Say Fqk is F_{(p^e)^2} = F_{p^e}[X] / (..).

An element `a + b X` of Fqk is called unitary if `a^2 + b^2 = 1`. Fqk elements in the verification key are unitary, which means the `b` coordinate can be deduced from the `a` coordinate up to a sign. So, we can compress `a + b X` down to `a` plus a bit. The bit that we use (because extracting it is convenient in the circuit) is the parity of the "real part" of `b`. But this bit is only enough information to figure out the sign if `b` has non-zero real-part. So, the real part of `b` should be nonzero and we have an assertion enforcing this.

A simple way of accomplishing this is making everything all ones, which is what I did here.

### DIFF
--- a/src/camlsnark_c/libsnark-caml/depends/libff/libff/algebra/fields/fp4.hpp
+++ b/src/camlsnark_c/libsnark-caml/depends/libff/libff/algebra/fields/fp4.hpp
@@ -67,6 +67,7 @@ public:
     Fp4_model cyclotomic_squared() const;
 
     std::vector<my_Fp> all_base_field_elements() { return { c0.c0, c0.c1, c1.c0, c1.c1 }; }
+    Fp4_model(const std::vector<my_Fp>& v) : c0(my_Fp2(v[0], v[1])), c1(my_Fp2(v[2], v[3])) {};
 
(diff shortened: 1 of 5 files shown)


src/camlsnark_c/libsnark-caml/depends/libff/libff/algebra/fields/fp4.hpp
```diff
@@ -67,6 +67,7 @@ class Fp4_model {
     Fp4_model cyclotomic_squared() const;
 
     std::vector<my_Fp> all_base_field_elements() { return { c0.c0, c0.c1, c1.c0, c1.c1 }; }
+    Fp4_model(const std::vector<my_Fp>& v) : c0(my_Fp2(v[0], v[1])), c1(my_Fp2(v[2], v[3])) {};
```
mrmr1993: Nit: can you move this next to the other `Fp4_model` declarations further up the class.

PR #348, opened 2019-08-06
Throw more informative errors when checked computations raise exceptions

mrmr1993:


src/checked_runner.ml
```diff
@@ -318,6 +318,16 @@ module Make (Backend : Backend_extended.S) = struct
         let s, y = next_auxiliary s in
         run (k y) s
 
+  let run t s =
+    try run t s
+    with exn ->
+      Printf.eprintf
```
imeckler: Hmm I am a bit skeptical about just writing to stderr. Maybe this should return a more structured error object and the caller can print?

mrmr1993:
Updated. This now
* raises a `Snarky.Checked_runner.Runtime_error` (aliased in `Snarky.Snark0`/`Snarky.Snark`), including
  - a formatted message that can be shown to the user (including the original error, the snarky backtrace and the OCaml backtrace)
  - the relevant raw information from the original error: the snarky label stack, the exception, and the backtrace from the original exception (up to where we caught it)
* adds a printer to `Printexn`, so that the output of the exception is better integrated with the normal OCaml error text
* moves the `try .. with ...` into the original `run` function, so that it can see the actual label stack at the error (before it was just finding the initial stack.. which is always empty :sweat_smile:)

An error in a checked computation now looks like this:
```
Uncaught exception:

  Snarky.Checked_runner.Runtime_error(_, _, _, _)

Encountered an error while evaluating the checked computation:
  (Failure "oh dear")

Label stack trace:
winner: File "examples/election/election.ml", line 96, characters 0-589
check_winner: File "examples/election/election.ml", line 111, characters 0-129


Raised at file "stdlib.ml", line 33, characters 17-33
Called from file "examples/election/election.ml", line 106, characters 15-33
Called from file "src/checked_runner.ml", line 317, characters 14-19


Raised at file "src/checked_runner.ml", line 348, characters 8-55
Called from file "src/snark0.ml", line 141, characters 13-26
Called from file "src/snark0.ml" (inlined), line 961, characters 29-59
Called from file "src/snark0.ml", line 969, characters 23-59
Called from file "examples/election/election.ml", line 120, characters 14-66
```

PR #358, opened 2019-08-22
delete curves.ml and fix Triple.map

imeckler:
This PR

- Removes `curves.ml`
- Fixes the buggy implementation of `Triple.map` and adds mlis in tuple_lib


tuple_lib/triple.ml
```diff
@@ -2,4 +2,4 @@ open Core_kernel
 
 type 'a t = 'a * 'a * 'a [@@deriving bin_io, sexp, eq, compare]
 
-let map (x1, x2, x3) ~f = (f x1, f x2, x3)
+let map (x1, x2, x3) ~f = (f x1, f x2, f x3)
```
mrmr1993: 😓

PR #367, opened 2019-09-11
[Meja] Substitutions

mrmr1993:
This PR implements a `Subst` module, used to rename modules and type names within types.

This results in
* improved error messages (with correctly scoped types)
* type names that always point to a type that is findable in the current environment
  - this is important for reconstructing types from a serialised typed AST
  - this also lets us remove the `var_decl` field from `Type0.variant`, which makes some `Type0.type_decl`s recursive
    + as a bonus, this will lets us implement mutable unification, which vastly simplifies type stitching and the unification algorithms and environment in general
* the possibility of implementing functors properly


meja/tests/module-scoping.stderr
```diff
@@ -0,0 +1,4 @@
+File "tests/module-scoping.meja", line 33, characters 14-15:
+Error: Incompatable types B.t -> B.A.t -> B.B.t -> _ and
```
psteckler: "Incompatible"

PR #382, opened 2019-09-18
[Meja] Use paths to resolve type declarations

mrmr1993:
This PR
* removes `var_decl` from `variant`, instead using the path in `var_ident` to find the type declaration in the environment
  - this makes it much easier to stitch type declarations; without it, we need to stitch all of the types in both of the type declarations, some of them including the type declarations themselves
* makes some tweaks to path resolution to ensure that the paths stay up-to-date (and so we can always find the type declaration!)
  - part of this involved changing when modules could be deferred: previously *any* module from within an external file could be deferred, which would lead substitution to do the wrong thing; now only the module representing the entire external library can be deferred
  - deferred modules were also simply *assumed* to exist (and to be in `Checked` mode), which could have lead to nonsense code that typechecks
  - constructor resolution also had a few edge cases where the path may not resolve properly; these are now fixed
* removes `path` from `Envi.Scope.t` entirely
  - if we use the scope's path instead of a path from the AST or a type (which `subst` may have changed) then we may try to look up type declarations in the wrong place
  - this wasn't used anywhere where we didn't have more reliable information anyway


meja/src/envi.ml
```diff
@@ -519,6 +499,56 @@ module Scope = struct
...
+      the loading of the module at their creation time.
+  *)
+  let get_module_by_ident ~loc resolve_env (ident : Ident.t) (scope : t) =
+    match IdTbl.find ident scope.modules with
+    | Some (Immediate m) ->
+        Some m
+    | Some (Deferred path) -> (
+      (* Use Prover mode so that we pick up all the modules. *)
+      match get_global_module ~mode:Prover ~loc resolve_env path with
+      | Some (_, m) ->
```
imeckler: Consider `Option.map ~f:snd`

meja/src/envi.ml
```diff
@@ -519,6 +499,56 @@ module Scope = struct
...
+      only be used to retrieve sub-values of the module, which must have forced
+      the loading of the module at their creation time.
+  *)
+  let get_module_no_load ~loc resolve_env ~mode name scope =
+    match IdTbl.find_name name ~modes:(modes_of_mode mode) scope.modules with
+    | Some (ident, Immediate m) ->
+        Some (ident, m)
+    | Some (ident, Deferred path) -> (
+      (* Use Prover mode so that we pick up all the modules. *)
+      match get_global_module ~mode:Prover ~loc resolve_env path with
```
imeckler: similar

PR #389, opened 2019-09-25
[Meja] Type stitching

mrmr1993:
This PR implements type stitching. Overview:
* there is a new constructor `Treplace` used for recursion breaking by `Type1.flatten`, `Envi.Type.copy` and `Type0_map.type_expr`, tracked by a new `Replace _ : Type1.change` in snapshots
  - this is functionally equivalent to `Tref` and the `Desc` change constructor, but separating it out lets us undo the recursion breaking without undoing unifications etc. that may also have happened
* `Type1.Snapshot` can now filter snapshots, only reversing the changes that match the filter. We use this to backtrack only the `Replace` changes when we've used them for recursion breaking
* adds modes to types with a `type_mode` field
* adds recursion protection to `typ_debug_print`
  - with `Treplace` and the way it is used in `Type0_map.type_expr`, the types can be self-referential, and so trying to print them lead to a stack overflow. It is now possible to (and was really helpful to) print types in these states
  - some of the printing was also tweaked to make it more compact (paths/idents) and to display the type's mode
* adds a `type_alternate` field to `Type0.type_expr`, representing the type to which the current type is stitched
  - ensuring that this is kept consistent is the bulk of this PR, and motivated all of the other changes

**NOTE: This doesn't stitch type declarations, which will be done instead in a follow-up PR.** Those changes should be comparatively straightforward, with the main bulk being just looking up the alternate when instantiating type constructors, but there are enough delicate things going on here that it seems worth separating.


mrmr1993:
@imeckler Woops, managed to miss this review, sorry! There's nothing user-facing here to test yet, but the `assert` sanity checks in this PR make sure that everything stays sane in all of our existing tests.

I'm still working on the type declaration stitching ('unconvertable' types, where we should be able to go `Checked -> Prover` but not `Prover -> Checked`, are more complicated than I expected), but I'm integrating that with our codegen so I'll be able to add some nice tests there.

PR #394, opened 2019-10-02
[Meja] Add a dedicated conversion type

mrmr1993:
Note: this PR includes #389.

This PR adds a new type kind for conversions.
* The type `'a --> 'b` represents `('a, 'b) Typ.t` where `'a` is interpreted in checked mode and `'b` is interpreted in prover mode.
  - statements of this type will be generated in the course of generating stitched type declarations
  - signature elements of this type can be described, so that we can expose the snarky internal instances such as `(Boolean.t, bool) Typ.t` and `(Field.Var.t, Field.t) Typ.t` easily
* Internally, this is represented as `Tconv _`, where `_` is a type comprised of the stitched types between which it represents the conversion
  - this makes it trivial to check for correct instances for conversions: we simply check t the type of conversions between `'a` and `'b` hat both parts of a stitched type unify with the corresponding parts of the stitched type argument to `Tconv`
* This type kind is the kind used to annotate the conversion AST instances introduced in #393.


mrmr1993:
Closing, this was included in #418.

PR #396, opened 2019-10-14
[Meja] Fix recursive type declaration imports

mrmr1993:
This PR adds support for recursive type declaration imports from OCaml `.cmi` files.


mrmr1993:
Updated. This now goes much further than the original scope of the PR:
* there is now support for co-recursive types in statements and signatures
  - syntax `type rec t = ... and type u = ...;`
* the default for all types is that they are now non-recursive
  - in theory, we should therefore be able to synthesise a `Typ.t` for any normal type. In particular, we won't have trouble with self-recursive records any more.
* imported OCaml interfaces use the correct form of recursion for the type (so we don't end up with nonsense like `type t = t` when it should have been `type nonrec t = t`)

The predeclaration mechanism is separate from the rest of the environment, so it can be extended with the name of the stitched type, which gives us a sane way to generate the definitions separately but still ensure that they are correctly stitched.

PR #398, opened 2019-10-21
Add Typ.Internal.fn for converting functions in the imperative API

mrmr1993:
This `Typ.t` describes how to convert between checked functions and
prover functions, using a `Typ.t` for each of the argument and the
return type.

The implementation is 'unsafe', but when used with the API in Snark_intf
it should function correctly. There is an additional check that we are
in prover mode, so that explicitly calling [Typ_monads.Read.read] fails
instead of generating a witness for the incorrect R1CS.

----

This is necessary to give Meja a way to call 'checked' functions from prover mode, since the majority of definitions will live in the checked namespace. For example
```reason
open Field;

let ( * ) = fun ((x1, x2) : field * field, (y1, y2) : field * field) => {
  (x1 * y1, x2 * y2);
};

Prover {
  let square x = x * x;
};

/* Evaluate square, using a square constraint instead of a normal multiplication constraint. */
let square x =
  let y = Prover { square x; };
  let ((x1, x2), (y1, y2)) = (x, y);
  assert_square x1 y1 ;
  assert_square x2 y2 ;
  y
```

More generally, it gives us a simple trapdoor to evaluate a checked-namespace function in prover mode where we don't need a type conversion, e.g. `Prover {if (a) {b && c;} else {c || d;};}` (where `&& : bool -> bool -> bool` and `|| : bool -> bool -> bool` live in the checked namespace to ensure they are accessible to both checked and prover code).

This also gives a reasonable way to surface locally shared 'prover' functions within a checked function (unless we decide to make `Prover { ... } : _ -> _` an error):
```reason
let f = fun () => {
  let file = Prover { open_file("filename"); };
  let read = Prover { read_one_from_file; };
  let x = read(file);
  let y = read(file);
  assert_r1cs(x, y, 1f);
}
```


imeckler:
Is it appropriate to use Typ.t for this purpose? I think for now fine but maybe there should be some iso type (does this have any relationship to https://github.com/o1-labs/snarky/pull/394 ?)

mrmr1993:
> Is it appropriate to use Typ.t for this purpose?

It's not ideal, but it lets us work around weird situations with tuples etc. so that prover mode can handle e.g. a value `: (field, field -> field)` without having to hack in an extra value-dependency-analysis algorithm or giving a nasty error message.

PR #399, opened 2019-10-23
Snarky universe

imeckler:
This creates the `snarky_universe` library which provides an easy to use snarky module. Right now it has support for

- field elements
- bools
- hashing
- merkle trees

It is intended to match the interface of [this library](https://github.com/o1-labs/snarkyjs-crypto). I also added a "anonvote" example which uses this library.


examples/anonvote/anonvote.ignore.ml
```diff
@@ -0,0 +1,73 @@
...
+
+let attributeCount = 10
+
+let merkleTreeDepth = 25
+
+module Voter = struct
+  type ('priv_key, 'attr) t_ = {privateKey: 'priv_key; attributes: 'attr array}
+  [@@deriving yojson]
+
+  (* TODO: Make this hash *)
```
mrmr1993: We should do this TODO

examples/anonvote/anonvote.re
```diff
@@ -0,0 +1,100 @@
...
+let merkleTreeDepth = 25;
+
+module Voter = {
+  [@deriving yojson]
+  type t_('priv_key, 'attr) = {
+    privateKey: 'priv_key,
+    attributes: array('attr),
+  };
+
+  /* TODO: Make this hash */
```
mrmr1993: Ditto

snarky_universe/hash.ml
```diff
@@ -0,0 +1,128 @@
...
+  open Impl
+
+  module Inputs = struct
+    include (val Poseidion_rounds.of_alpha (alpha_of_curve C.curve))
+
+    module Field = struct
+      (* The linear combinations involved in computing Poseidon do not involve very many
+    variables, but if they are represented as arithmetic expressions (that is, "Cvars"
+    which is what Field.t is under the hood) the expressions grow exponentially in
+    in the number of rounds. Thus, we compute with Field elements represented by
```
mrmr1993: Perhaps not immediately, but we should add something to deal with this better in general.

sponge/params.sage
```diff
@@ -0,0 +1,92 @@
...
+            return mds
+
+for name, r in [  ('Bn128', BN128_r), ('Mnt4_298', MNT4298_r), ('Mnt4_753', MNT4753_r) ]:
+  F = FiniteField(r)
+  print ('let params_{} = '.format(name)
+          + '{ Params.mds=' + matrix_str(mds(F)) + ';'
+          + 'round_constants= ' + matrix_str(round_constants(F))
+          + '}' )
+
+#F_small= FiniteField(BN128)
```
mrmr1993: Can we delete these?

sponge/sponge.ml
```diff
@@ -0,0 +1,136 @@
...
+  let copy = Array.copy
+end
+
+let m = 3
+
+module Rescue (Inputs : Intf.Inputs.Rescue) = struct
+  (*
+   We refer below to this paper: https://eprint.iacr.org/2019/426.pdf.
+
+I arrived at this value for the number of rounds in the following way.
```
mrmr1993: Nice, thank you for this!

snarky_universe/snarky_universe.ml
```diff
@@ -0,0 +1,271 @@
...
+            type t = unit
+          end)
+      in
+      (module Run : Snarky.Snark_intf.Run with type prover_state = unit)
+
+module Make (C : sig
+  val curve : Curve.t
+end)
+() : Intf.S = struct
+  module Impl = (val impl_of_curve C.curve)
```
mrmr1993: I think `struct .. end` is needed here.

imeckler: functors can take a "unit" argument like this to make them generative rather than applicative. Basically applicative means you get type equalities between different applications whereas generative means you don't.

PR #400, opened 2019-10-25
Unchecked hash and merkle tree + integers

imeckler:
This adds 

- integers to snarky universe (just copying over snarky_integer from coda)
- an unchecked hash and Merkle tree, both copied over from snarky-cryptojs

### DIFF
--- a/snarky_universe/snarky_universe.ml
+++ b/snarky_universe/snarky_universe.ml
@@ -1,3 +1,5 @@
+module B = Bigint
+
 module Cond (Impl : sig
   module Boolean : sig
     type var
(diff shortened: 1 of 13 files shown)


snarky_universe/snarky_universe.ml
```diff
@@ -90,6 +92,10 @@ end)
 
     let assertEqual (x : t) (y : t) = Field.(Assert.equal (x :> t) (y :> t))
 
+    let assertTrue = assertEqual true_
```
mrmr1993: Nice!

PR #402, opened 2019-10-25
examples for SFBW and some fun(ctions)

imeckler:
I apologize for the ill-defined scope of this PR. It includes:

- several new examples for the SFBW workshop
- some functions for checking if field elements are square and taking square roots which I took from coda.


examples/sfbw/ex00_preimage/ex00_preimage.re
```diff
@@ -0,0 +1,8 @@
+module Universe = (val Snarky_universe.default());
```
mrmr1993: It looks like these examples build executables that do nothing

imeckler: I'm going to integrate them with the json top level functor once they're both in. I'm also making a few changes on your branch to the functor if that's alright with you.

mrmr1993: Sounds good. Will approve this now then, so you can merge later!

PR #406, opened 2019-10-28
Add support for prover state to js_snarky

mrmr1993:
This PR adds direct support for `prover_state` to `js_snarky`. This mean that
* the prover can be passed data without needing to expose it directly in the R1CS
* the hack for working around this limitation can be removed.


imeckler:
Sorry just seeing this -- do you mind if we leave the hack in for the workshop to minimize changes in the examples I have in my branch?

mrmr1993:
Sure thing; just threw this out there in case it made it easier to do some more data-intensive examples!

PR #409, opened 2019-10-29
Store R1CS hash with keys by default, re-generate when reading

mrmr1993:
This PR tweaks the `Proof_system` interface to
* by default, store the r1cs hash with the proving key
* reject the proving key if the r1cs hash doesn't match
  - this causes a fresh key to be generated

This also tidies up the `snarkyjs` node command, so the names match the `js_snarky` -> `snarkyjs` change, and so that the default set-up steps work again with:
```bash
sudo npm install -g path/to/snarkyjs
cd new/project/directory
snarkyjs init ; npm link snarkyjs
node index.js
```

### DIFF
--- a/snarkyjs/index.js
+++ b/snarkyjs/index.js
@@ -78,9 +78,30 @@ module.exports = exports = function (name) {
         });
       }, function (err) { return err; });
   };
+  var generate_keys = function() {
+    var query_json = JSON.stringify({"command": "verify"});
+    return communicate(query_json).then(function (response_json) {
+        return new Promise (function (resolve, reject) {
+          var response = JSON.parse(response_json);
+          if (response.name == "error") {
+            reject(new Error(response.message));
+          } else if (response.name == "keys_generated") {
+            resolve({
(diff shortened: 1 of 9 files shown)


imeckler:
Does this also generate keys at startup? @nholland94 mentioned he needs that

mrmr1993:
Updated: added a `generate_keys` command. It doesn't automatically get run at startup -- presumably we don't want to clobber existing keys if they have been deliberately placed there etc. -- but you can call it straight after creating a `new Snarky("path/to/exe.exe")` to achieve the same effect.

snarkyjs/index.js
```diff
@@ -78,9 +78,30 @@ module.exports = exports = function (name) {
         });
       }, function (err) { return err; });
   };
+  var generate_keys = function() {
+    var query_json = JSON.stringify({"command": "verify"});
```
nholland94: Pretty sure the command here should be "generate_keys".

mrmr1993: Good catch! Fixed.

PR #414, opened 2019-10-31
[Meja] Scoped implicits

mrmr1993:
This PR
* changes the instance resolution algorithm to accept the first instance found, instead of failing with an error
* makes instances scope local
  - this ensures that the paths to instances always resolve correctly
* adds an `open implicit M` statement for any module `M`
  - this allows us to include the instances from an external module even when it isn't otherwise referred to, and without having to open it when we don't necessarily want it open
  - this also gives us an easy way to override instances by including a module with some different instances
* adds a expression-level `instance`
  - this was the main motivation, for use in the `Typ.t` generation
  - this will also allow us to create and instantiate expression-local request handlers when row-subtyping is (finally) implemented

This was originally part of the ongoing tri-stitching/`Typ.t` generation typechecker overhaul, but this is separate enough to break out as its own PR.

### DIFF
--- a/meja/ocaml/of_typedast.ml
+++ b/meja/ocaml/of_typedast.ml
@@ -418,6 +418,12 @@ let rec of_expression_desc ?loc = function
       Exp.let_ ?loc Nonrecursive
         [Vb.mk (of_pattern p) (of_expression e_rhs)]
         (of_expression e)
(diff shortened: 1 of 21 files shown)


mrmr1993:
Closing, this was included in #418.

PR #421, opened 2019-11-11
[Meja] Generate unique type variable names for printing

mrmr1993:
This PR
* adds a mechanism for generating distinct type variable names
* uses this mechanism to assign distinct type variable names to each type in `Typechecker` error messages
  - This makes several ambiguous/unclear error messages clearer, most notably `tests/recursive-type.stderr`

This will also form a necessary part of printing out module signatures when performing multi-file builds: we will need to functor the module over the snarky interface *and the signatures of all modules on which it depends* in order to allow switching between fields and proof systems easily, and so we need to be able to print signatures including type variables unambiguously.

### DIFF
--- a/meja/src/envi.ml
+++ b/meja/src/envi.ml
@@ -634,6 +634,57 @@ let map_current_scope ~f env =
   | [] ->
       raise (Error (of_prim __POS__, No_open_scopes))
 
+let get_var_names env =
+  let rec go map set scope_stack =
+    match scope_stack with
+    | ({Scope.kind= Expr; _} as current_scope) :: scope_stack ->
+        let map, set =
+          Map.fold ~init:(map, set) current_scope.type_variables
+            ~f:(fun ~key ~data (map, set) ->
+              (Map.add_exn map ~key:data ~data:key, Set.add set key) )
+        in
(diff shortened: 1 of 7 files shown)


meja/src/envi.ml
```diff
@@ -634,6 +634,57 @@ let map_current_scope ~f env =
   | [] ->
       raise (Error (of_prim __POS__, No_open_scopes))
 
+let get_var_names env =
+  let rec go map set scope_stack =
+    match scope_stack with
```
imeckler: Nit : this could be a fold

meja/src/envi.ml
```diff
@@ -634,6 +634,57 @@ let map_current_scope ~f env =
...
+
+let set_var_names env =
+  let typ_map, name_set = get_var_names env in
+  let typ_map = ref typ_map in
+  let name_set = ref name_set in
+  let next_name = ref 0 in
+  let get_next_name () =
+    let id = !next_name in
+    incr next_name ;
+    if id < 26 then String.of_char (Char.of_int_exn (Char.to_int 'a' + id))
```
imeckler: Nice

PR #426, opened 2019-11-13
Update libsnark, libff, and libfqfft

enolan:
This may still be broken on Mac, due to exceptions in constructors.


imeckler:
How should we make sure it works on Mac?

PR #433, opened 2019-11-27
[Meja] Row types

mrmr1993:
This PR is a work-in-progress.

Currently, it implements row types in the typechecker. These are compatible with OCaml's row types, but the implementation is more flexible on the constructors (an identifier/path pair instead of a string), so that it is possible to use them for effects too.

~~Row type expressions and patterns are not yet implemented.~~


mrmr1993:
Row constructors in expressions and patterns are now implemented.

meja/tests/row-types.meja
```diff
@@ -0,0 +1,368 @@
...
+
+  let args : [<`A(int) | `B(bool)];
+
+  let arg_noarg : [<`A(int) | `B];
+
+  let noarg_arg : [<`A | `B(bool)];
+};
+
+module type Bounded_rows = {
+  let fixed_no_args : [< `A | `B > `A | `B ];
```
imeckler: What is the meaning of this kind of type? Do you think you could put a spec somewhere? 

Also - do you think you could get requests working in an untyped way (like we have with the library) to get to a working version sooner or is it somehow entangled?

mrmr1993: Somehow managed to miss this review, sorry.

This type is analogous to the OCaml type of the same form:
```ocaml
[< `Tag_1 | `Tag_2 | `Tag_3 | ... | `Tag_n > `Tag_i | `Tag_j | ... | `Tag_k]
```
is a type that may only be constructed by the constructors to the left of `<`, and the minimum set of constructors are those to the right of `>`. For example,
```ocaml
val f : [< `A | `B | `C > `A] -> _

(* This typechecks, restricting the type to the minimum set of constructors. *)
let g (x : [`A]) = f x

(* This typechecks, all of the constructors are present. *)
let h (x : [`A | `B | `C]) = f x
```



> Also - do you think you could get requests working in an untyped way (like we have with the library) to get to a working version sooner or is it somehow entangled?

Requests should currently work, but we have to attach handlers manually with `handle`, as in OCaml. The row typing work is to infer a) when we have a handler that we can attach, so that we can do so manually and b) when we try to compile a whole program but there is no handler for a request that isn't hit.

mrmr1993:
This was merged as part of #442. Closing.

PR #445, opened 2020-01-03
two merkle tree modifications per request

deepthiskumar:
...without computing the hashes for the intermediate root

### DIFF
--- a/src/merkle_tree.ml
+++ b/src/merkle_tree.ml
@@ -453,11 +453,57 @@ struct
     let%map new_root = implied_root next_entry_hash addr0 prev_path in
     (new_root, prev)
 
+  (*without computing the  intermediate root*)
+  let%snarkydef_ pair_of_fetch_and_update_req ~(depth : int) root
+      (addr1, addr2) ~f : (Hash.var, 's) Checked.t =
+    let open Let_syntax in
+    let update_tree addr elt =
+      perform
+        (let open As_prover in
+        let open Let_syntax in
+        let%map addr = read (Address.typ ~depth) addr
+        and next = read Elt.typ elt in
+        Set (addr, next))
+    in
+    let%bind prev1, prev_path1 =
+      request_witness
+        Typ.(Elt.typ * Path.typ ~depth)
+        As_prover.(
+          read (Address.typ ~depth) addr1 >>| fun addr -> Get_element addr)
+    in
+    let%bind prev2, _prev_path2 =
+      request_witness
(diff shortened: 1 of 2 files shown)


src/merkle_tree.ml
```diff
@@ -453,12 +453,58 @@ struct
...
+    let%bind () = update_tree addr1 next1 in
+    let%bind () = update_tree addr2 next2 in
+    let%bind next_elt_hash = Elt.hash next2 in
+    (*TODO: generate the path while setting the new value*)
+    let%bind new_path2 =
+      request_witness (Path.typ ~depth)
+        As_prover.(
+          read (Address.typ ~depth) addr2 >>| fun addr -> Get_path addr)
+    in
+    let%map new_root = implied_root next_elt_hash addr2 new_path2 in
```
mrmr1993: If this is only dependent on `new_path2`, `addr2` and `next_elt_hash`, it seems to me like the merkle tree can be arbitrary (except for having `next_elt` at `addr2`) and still satisfy the R1CS. Is there something I'm missing/could you talk to why this is okay?

mrmr1993: Also nit: the `let%map new_root = ... in new_root` here can be omitted.

deepthiskumar: What I was intending to do: 
1. get elt1 and elt2 at addr1 and addr2
2. apply f and get next1 and next2
3. update element at  addr1 with next1
4. get the new path (list of hashes) for elt2 at addr2
5. check if the root at the new path is the same as the root after updating the first element (this the step I thought could be skipped). This would ensure the two elements are from the same merkle tree
6. update elt2 with next2 and return the root after the update

But as @mrmr1993 pointed out this wouldn't satisfy the RICS. Made changes at the call site to use two separate `fetch_and_update` requests

src/merkle_tree.ml
```diff
@@ -451,59 +451,13 @@ struct
...
-    let%bind next_elt_hash = Elt.hash next2 in
-    (*TODO: generate the path while setting the new value*)
-    let%bind new_path2 =
-      request_witness (Path.typ ~depth)
-        As_prover.(
-          read (Address.typ ~depth) addr2 >>| fun addr -> Get_path addr)
-    in
-    let%map new_root = implied_root next_elt_hash addr2 new_path2 in
-    new_root
+    (new_root, prev, next)
```
mrmr1993: It would be good to mark these values so that the old and new value are obvious. Perhaps something like

```ocaml
(new_root, `Old prev, `New next)
```

PR #451, opened 2020-01-09
Pare down backend

imeckler:
This PR prepares snarky to work with the new curve backends.

Previously backend_extended had a bunch of code which used libsnark specific functionality to implement a few functions on constraint systems (e.g., computing the digest) Now, a backend is expected to provide those functions itself. A functor has been added for generating those functions from a libsnark backend.

Also, in the new backend, the constraint system is not a pointer but an ocaml value. As such, instead of the `run_state` having a pointer in it, I have replaced it with a "poor man's object" of a constraint system. That is, an abstractly typed value paired with a module with the necessary functions on that type (`add_constraint`, `set_primary_input_size`, etc.)

### DIFF
--- a/src/camlsnark_c/bindings/vector.mli
+++ b/src/camlsnark_c/bindings/vector.mli
@@ -1,4 +1,4 @@
-type 'a t = unit Ctypes.ptr
+type 'a t
 
 val null : 'a t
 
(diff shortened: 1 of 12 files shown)


src/camlsnark_c/bindings/vector.mli
```diff
@@ -1,4 +1,4 @@
-type 'a t
+type 'a t = unit Ctypes.ptr
```
mrmr1993: This makes me a little uncomfortable, since it makes all of the vector types alias to one another. Would it be a problem to make this opaque again?

PR #466, opened 2020-04-10
Use libsodium for randomness

imeckler:
This PR modifies libff to use sodium for generating random field elements. This fixes a potential vulnerability on certain old platforms


ValarDragon:
In libiop, libsodium calls for O(N) random bits was a notable slow-down. I'm not sure if that is the case for your zk-polynomial commitment.

If it is, one thing we did to fix this was to set some batch size, and get randomness in larger chunks than just 1 field elem.

PR #467, opened 2020-04-11
Add daira's fast scalar multiplication

imeckler:
This PR implements Daira's fast scalar multiplication algorithm from [here](https://github.com/zcash/zcash/issues/3924).

### DIFF
--- a/snarky_curve/snarky_curve.ml
+++ b/snarky_curve/snarky_curve.ml
@@ -307,10 +307,10 @@ module type Native_base_field_inputs = sig
 end
 
 module For_native_base_field (Inputs : Native_base_field_inputs) = struct
-  include Make_checked (Inputs)
   open Core_kernel
   open Inputs
   open Impl
+  include Make_checked (Inputs)
 
   module Window_table = struct
     open Tuple_lib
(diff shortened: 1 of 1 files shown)


snarky_curve/snarky_curve.ml
```diff
@@ -447,4 +447,170 @@ module For_native_base_field (Inputs : Native_base_field_inputs) = struct
...
+
+     Output:
+    (2*r + 1 + 2^len(r)) t
+  *)
+  let scale (t : Field.t * Field.t) (`Times_two_plus_1_plus_2_to_len r) :
+      Field.t * Field.t =
+    let n = Array.length r in
+    let acc = ref (double t) in
+    let () =
+      for i = 0 to n - 1 do
```
daira, line 509: This is reversed relative to the https://github.com/zcash/zcash/issues/3924 algorithm; it's probably just that r is MSB-first, but that isn't documented.

(Same in the [current code](https://github.com/o1-labs/snarky/blob/2a85945bb4855484ba076257839bcf77bb65244c/snarky_curve/snarky_curve.ml#L528).)

PR #470, opened 2020-04-22
Statically link libsodium

psteckler:
Coda already statically links `libsodium`, so this change prevents executables from the coda repo built with snarky trying to dynamically load a `libsodium.so`.

### DIFF
--- a/src/camlsnark_c/camlsnark_linker_flags_gen.ml
+++ b/src/camlsnark_c/camlsnark_linker_flags_gen.ml
@@ -22,7 +22,9 @@ let () =
         ; "-lgomp"
         ; "-lssl"
         ; "-lcrypto"
-        ; "-lsodium"
+        ; "-Wl,--push-state,-Bstatic"
+        ;  "-lsodium"
+        ; "-Wl,--pop-state"
         ; "-lprocps"
         ; "-lgmp"
         ; "-lstdc++" ]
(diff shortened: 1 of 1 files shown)


src/camlsnark_c/camlsnark_linker_flags_gen.ml
```diff
@@ -22,7 +22,9 @@ let () =
         ; "-lgomp"
         ; "-lssl"
         ; "-lcrypto"
-        ; "-lsodium"
+        ; "-Wl,--push-state,-Bstatic"
+        ;  "-lsodium"
+        ; "-Wl,--pop-state"
```
mrmr1993, line 27: We dynamically link `libsodium` on Macs in coda too, right?

mrmr1993: Indeed, this is the case.

psteckler: right, see `src/external/ocaml-sodium/myocamlbuild.ml` in coda.

psteckler: https://github.com/CodaProtocol/coda/blob/develop/src/external/ocaml-sodium/myocamlbuild.ml#L11

PR #471, opened 2020-04-22
Move in groupmap and sponge

imeckler:
This PR

- moves the group map library into snarky and uses it to produce an "unrelated base" for the purposes of the optimized in-snark exponentiation
- adds the bn382 sponge inputs to `sponge`

### DIFF
--- a/group_map.opam
+++ b/group_map.opam
@@ -1,0 +1,6 @@
+opam-version: "1.2"
+version: "0.1"
+build: [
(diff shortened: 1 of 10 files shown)


group_map.opam
```diff
@@ -0,0 +1,6 @@
+opam-version: "1.2"
```
mrmr1993, line 1: This needs to be a proper opam file, otherwise pinning from this repo will break

PR #477, opened 2020-05-15
Build requires jbuilder for meja and snarky_bench

pirapira:
Both changed opam files contain build commands using `jbuilder`.
It was missing from the list of dependencies.


mrmr1993:
`jbuilder` is now deprecated, so I'm hesitant to re-introduce it as a dependency. A PR to change the `jbuilder` calls to `dune` ones would be appreciated though!

pirapira:
I agree that's a better course of action.

PR #479, opened 2020-05-21
Remove a never-used functor Conv

pirapira:
This PR removes a functor Conv, which seems to be used nowhere.


mrmr1993:
This functor is intended for use by consumers of the library, so that they can use their own constraint format in their proof backends. Could do with a doc-comment though: would be great to see a PR to add one!

pirapira:
OK!

PR #480, opened 2020-05-22
Avoid mutable can_mutate_state

pirapira:
and instead use an immutable argument to `go` function.


pirapira:
I'm pulling the docker image to reproduce the build error.

pirapira:
Squash & force-push in order to start CicleCI. (Tests passed on CircleCI but GitHub was still waiting.)

PR #484, opened 2020-05-29
Replace jbuilder with dune in snarky_bench build

pirapira:
as suggested in #477.


pirapira:
I wrote first `rune` instead of `dune` everywhere, and fixed it in-place in this PR. However, I could not fix the branch name.

PR #496, opened 2020-06-30
Sponge updates

imeckler:
This PR

- adds sponge constants for the tweedle curves
- exposes some more types from the sponge module


sponge/sponge.ml
```diff
@@ -219,17 +223,25 @@ module Make_hash (P : Intf.Permutation) = struct
...
+let make ~state ~params ~sponge_state = {state; params; sponge_state}
+
 module Make_sponge (P : Intf.Permutation) = struct
   open P
 
+  let make = make
+
   let capacity = 1
 
   type sponge_state = Absorbed of int | Squeezed of int [@@deriving sexp]
```
mrmr1993, line 242: Nit: I think this needs to be aliased to the new `sponge_state` type above, or deleted.

PR #499, opened 2020-07-01
Remove fast project

imeckler:
This stops using the fast project function as it was incompatible with the zexe backends. @mrmr1993 do you think I should just remove the commented out unit test or is it worth keeping it in the file?


mrmr1993:
> do you think I should just remove the commented out unit test or is it worth keeping it in the file?

Nah, kill it!

PR #500, opened 2020-07-08
Update snarkette bin_io

imeckler:
This updates the `snarkette` bin_io for field elements to be compatible with bin_io in the new backends. The main difference in that previously we serialized the number of bytes in the field element and now we don't.


mrmr1993:
Not merging this until we're ready to land the new snark in coda, so that we won't need a separate branch for other snarky changes in the meantime

PR #501, opened 2020-07-09
Add a H_list ppx

mrmr1993:
This PR
* moves `H_list` to its own library
* adds `h_list.ppx` deriver, with `[@@deriving hlist`/`to_hlist`/`of_hlist]`.


mrmr1993:
Updated. This now includes some minor Meja changes:
* the `H_list` module is added to the paths available for meja, so that snarky modules that refer to it don't fail
* paths of the form `A.B.C` only immediately resolve `A.B` in `find_module_deferred`; the resolved module is used to find `C` which may or may not be deferred. This was necessary because
  - the new `Snarky` toplevel module refers to every module in snarky
  - dune rewrites each module in this internally to point to a `Snarky__` toplevel module (e.g. `Snarky__.As_prover` in place of `As_prover`)
  - meja was seeing the `Snarky__.As_prover` etc. and triggering an immediate load of the modules
  - some of the modules contained OCaml code that meja doesn't support (e.g. packed modules) and the compiler bailed

PR #504, opened 2020-07-27
Don't prescribe digestif.ocaml for Snarky_curve

mrmr1993:
This lets the compiled binary choose the digestif backend to use, which allows us to default to `digestif.c` and fall back to `digestif.ocaml` for JS in CodaProtocol/coda#5491.

Edit: This also bumps the opam deps to match CodaProtocol/coda#5505.


imeckler:
@mrmr1993 this seems to have broken running unit tests in snarky_curves. In the coda repo, running `./scripts/skip_if_only_frontend_or_rfcs.sh bash -c 'source ~/.profile && make build && (dune runtest src/lib --verbose --profile=dev -j8 || (./scripts/link-coredumps.sh && false))'` fails at snarky_curve with


```
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Digestif referenced from src/lib/snarky/snarky_curve/snarky_curve.cmxa(Snarky_curve)
```

PR #514, opened 2020-08-23
libff: remove "depends/" prefix from include directives

mobileink:
The "depends/" prefix makes the embedded libff dependent on the build structure of snarky.  This patch removes that dependency so that libff builds independently.  Note that test `/libff/algebra/curves/tests/test_bilinearity.cpp` will not link until PR #511  is merged.


mrmr1993:
I'm not sure that this is peeking out of the libff directory; see [here](https://github.com/scipr-lab/libff/blob/master/libff/algebra/curves/bn128/bn128_g1.hpp#L12) for the upstream version. Going to leave this open in case I've misunderstood though.

mobileink:
Yeah, upstream is broken too.  ate-pairing is not part of libff. This include path forces users to support a particular build structure, namely one involving a "depends" subdir.  That prevents, for example, using ate-pairing and libff as distinct Bazel external repositories.

This is fixed in the obazl upstream libff repo.

mrmr1993 (approved):
Makes sense, thanks

PR #518, opened 2020-08-24
camlsnark_c: fix bash build cmds on macos

mobileink:
Fixes a bug in the embedded bash build scripts in `snarky/src/camlsnark_c/snark_caml_*/dune` that prevents compile on MacOS - causes libtool to choke on popd because the prev line ends with \\ (continuation)


mrmr1993 (approved):
Good catch, thanks.

PR #520, opened 2020-08-24
Add Bazel support: xbyak

mobileink:
Initial Bazel PR.  To test:  `$ bazel test @xbyak//test`

NOTE: all tests pass on Linux, but one (test:jmp) fails on MacOS.

PRs to follow will add support per-external repo: ate-pairing, libff, libfqfft, libsnark, all of which are C/C++ only, then finally snarky.


WORKSPACE
```diff
@@ -0,0 +1,25 @@
...
+#     urls = ["https://github.com/obazl/xbyak/archive/bzl-1.0.tar.gz"],
+#     strip_prefix = "xbyak-bzl-1.0",
+#     # sha256 = 
+# )
+
+##########################################
+######## Non-bazel external repos ########
+
+## boost needed by: @xbyak//sample:calc
+git_repository(
```
mrmr1993, line 18: Am I right in thinking that builds will need an internet connection for this to work properly? It might be better to add this as a submodule and use a `path` directive instead.

WORKSPACE
```diff
@@ -0,0 +1,25 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+#################################
+#### Bazelized external libs ####
+local_repository( name = "xbyak" , path = "src/camlsnark_c/libsnark-caml/depends/xbyak" )
+# http_archive(
+#     name = "xbyak",
+#     urls = ["https://github.com/obazl/xbyak/archive/bzl-1.0.tar.gz"],
```
mrmr1993, line 9: Similar here: I think when it comes time for us to switch over, I would personally prefer that this be a submodule with a `path`.

bzl/tools/user.bazelrc
```diff
@@ -0,0 +1,24 @@
+## Override the WORKSPACE definitions of repos:
+build --override_repository=xbyak=/path/to/xbyak
```
mrmr1993, line 2: Shouldn't this `/path/to/xbyak` placeholder be filled in here?

bzl/tools/user.bazelrc
```diff
@@ -0,0 +1,24 @@
...
+build --color=yes
+
+# build --subcommands
+build --subcommands=pretty_print
+build --verbose_failures
+build --sandbox_debug
+
+# build --show_timestamps
+# build --keep_going
+build --jobs 600
```
mrmr1993, line 20: What's the relevance of this `600` here? A clarifying comment would be helpful.

WORKSPACE
mobileink: On the first point, see [Offline Builds](https://docs.bazel.build/versions/master/external.html#offline-builds).  On the second, see following comment.

WORKSPACE
mobileink: We can have both. In this case, `src/camlsnark_c/libsnark-caml/depends/xbyax` can be a submodule, or not, Bazel doesn't care.  So the migration path is probably: install Bazel support, with repos pointing to the embedded code. Then switch the embedded code to submodules; this will have no effect on Bazel.  Then, if we want to move to pure bazel, delete the submodules and depend only on the Bazel repos.

bzl/tools/user.bazelrc
mobileink: Sorry,I'll add some documentation.  This file is a template; to use it you would copy it to the root dir and replace /path/to/xbyak with the actual path.  The root .bazelrc file contains a line that loads user.bazelrc if it exists.

Based on my experience, I think using overrides in user.bazelrc is probably a better way of using embedded repo code, rather than using `local_repository` in the WORKSPACE. At least I'm leaning in that direction; both ways work.

bzl/tools/user.bazelrc
mobileink: Inertia.  That's what all the Bazel examples on the web have. ;)

WORKSPACE
mobileink: Also, note that we are mimicking a submodule by putting xbyak's WORKSPACE file in `src/camlsnark_c/libsnark-caml/depends/xbyak`.  Basically that gives us an embedded Bazel repo, since a WORKSPACE file determines a repo.  What that means is that you can only build the xbyak targets by using a repo-qualified label, e.g. @xbyak//test.  If you try `bazel test src/camlsnark_c/libsnark-caml/depends/xbyak/test`, Bazel will complain that you've crossed repo boundaries.  I'll add something to this effect to the documentation.

mrmr1993 (approved):

PR #522, opened 2020-08-25
xbyak bazel: put embedded xbyak in sync with o1-labs/xbyak repo

mobileink:
This makes the xbyak code embedded in snarky match the snarky branch of [o1-labs/xbyak](https://github.com/o1-labs/xbyak/tree/snarky).  The next step will be to replace this code with a submodule.


mobileink:
replace with a PR containing a single commit

PR #524, opened 2020-08-25
xbyak: use quotes not < > in include paths

mobileink:
Using < > for include paths causes headaches when the lib is used as  dependency because Bazel is strict about the distinction between quoted and system headers, plus it's technically incorrect, since < > are supposed to be for system headers.

### DIFF
--- a/src/camlsnark_c/libsnark-caml/depends/xbyak/sample/jmp_table.cpp
+++ b/src/camlsnark_c/libsnark-caml/depends/xbyak/sample/jmp_table.cpp
@@ -3,7 +3,7 @@
 */
 #include <stdio.h>
 #define XBYAK_NO_OP_NAMES
-#include <xbyak/xbyak.h>
+#include "xbyak/xbyak.h"
 
 const int expectTbl[] = {
 	5, 9, 12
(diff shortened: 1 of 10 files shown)


mrmr1993:
> Bazel is strict about the distinction between quoted and system headers, plus it's technically incorrect, since < > are supposed to be for system headers.

The only distinction the C and C++ specs make is that `<>` provides access to system headers, all other details of `<>` and `""` are implementation defined (except undefined behaviour for name collisions). Convention is
* `<>` provides acess to the stdlib (in an implementation defined manner), or consults the include directories otherwise
  + Accepted usage is for stdlib and external library headers
  + All of these are separate compilation units which link against `xbyak` as an external library, which matches this convention.
* `""` provides relative path access, falling back to searching the include directories
  + Accepted usage is for relative paths, where the header is library internal (ie. all names have internal linkage, or 'external' headers where the external library's headers does not expose them in an accessible way)

I am opposed to the build-system enforcing these kind of non-standard source code changes, and consider this a bug with Bazel. [This PR](https://github.com/bazelbuild/bazel/pull/10712) purposes to fix this issue, and suggests a workaround of manually using `-isystem`; I would be much happier with the workaround, and to advocate for that PR.

mobileink:
No prob.

mobileink:
Cancelling, C++ no longer used.

PR #536, opened 2020-10-24
Bazel initial commit PR: xbyak

mobileink:
This PR includes only changes in the xbyak embedded external dep. With one exception every change is a file add. The exception is a test file, #include directives changed to use quotes instead of <>.

Four copies of xbyak are included in the snarky source. The only one actually used by snarky is `src/camlsnark-c/libsnark-caml/depends/xbyak`.  The others are embedded within the other top-level direct deps like `src/camlsnark-c/libsnark-caml/depends/libff`.  They can be ignored, although they may contain changes to put them in sync with the toplevel copy.

To test:  `$ bazel test @xbyak//test`.  All but one test (`jmp`) should pass.


mobileink:
Cancelling, C++ version no longer used.

PR #552, opened 2020-12-10
Fix Frobenius coefficient of MNT4-753 Fq4

Pratyush:
This is calculated as `Fq2_qnr^((q^3 - 1) / 4)`.

The old one was incorrect.

(I don't know if this is useful for y'all, given that you're switching away from these curves, but thought I'd send a PR anyway)


mimoo:
thanks mate :D not useful anymore but thanks!

PR #561, opened 2021-04-19
Ppx for Typ.t parts

mrmr1993:
This PR adds a deriver to support the new `Typ.t` constructor added in #558.

This still has some quirks:
* this doesn't generate the `check` function yet
* the field type for signatures is always `Field.t`, even though it should be `Field.Var.t` or `Field.Constant.t` depending on which style of interface we're using

Tests are executed by running the ppx, checking it against the expected output (checked in). The output can be updated if the tests fail with `dune promote --root=.`


ppx/test/dune
```diff
@@ -0,0 +1,18 @@
...
+ (name run_ppx_snarky)
+ (libraries ppx_snarky))
+
+(rule
+ (targets typ_ppx.corrected)
+ (deps (:test testcases/typ_ppx.ml))
+ (action
+   (progn
+    (with-stdout-to typ_ppx.corrected (run ./run_ppx_snarky.exe %{test}))
+    (run ocamlformat -i typ_ppx.corrected))))
```
psteckler, line 11: so this is just verifying correct OCaml syntax?

psteckler (approved):
checked the output for the tests, looks fine

code looks fine

mrmr1993:
`[@deriving hlist]` has proved sufficient for our purposes. Closing

PR #562, opened 2021-05-03
[CI] adding workflow for building with different hosts and ocaml version

mimoo:


mimoo:
I think someone needs to enable Github actions for this repo for this PR's actions to show up, cc @mrmr1993

PR #565, opened 2021-05-07
removing unused proof API from snarky

mimoo:


imeckler:
Looks like you need to reformat with `make reformat` -- also build will probably fail on the examples

mimoo:
> Looks like you need to reformat with make reformat

what's the rationale behind `app/reformat` btw? Why not use the default formatting?

> build will probably fail on the examples

that is probably true, I'd be really good if we could get build to work in CI to test PRs like this one

mimoo:
rebasing

mimoo:
Looks like this: [Proof_system](https://github.com/o1-labs/snarky/blob/015426c2e1dddda5724644e21043c4c41c12d45c/src/toplevel.ml#L245-L245)

and this: [Proof_system](https://github.com/o1-labs/snarky/blob/015426c2e1dddda5724644e21043c4c41c12d45c/src/toplevel.ml#L418-L418)

are not building correctly. I could just delete these patterns from the match but is it really what we want to do here?

mimoo:
removing `src/toplevel.ml` after discussions

mrmr1993:
You'll also need to remove references to `Toplevel` elsewhere before this will compile.

mrmr1993 (approved):
Amazing! I've been looking forward to killing this for ages :D

mimoo:
looks like I can't merge this myself

PR #569, opened 2021-05-10
Remove libsnark

imeckler:
This PR removes libsnark and libsnark dependent code, which will be moved to a separate repository.


mimoo:
didn't pass CI because it was still trying to run the examples

PR #572, opened 2021-05-11
[CI] fixing dockerfile

mimoo:
The `ci/buildx` finally passes :)


.dockerignore
```diff
@@ -2,3 +2,4 @@
 _build
 *.install
 .git
+_opam
```
mimoo, line 5: for local switches

.github/workflows/build.yml
```diff
@@ -11,13 +11,7 @@ jobs:
         os:
           - macos-latest
           - ubuntu-latest
-          - windows-latest
         ocaml-version:
-          - 4.12.0
-          - 4.11.0
-          - 4.10.1
-          - 4.09.1
-          - 4.08.1
```
mimoo, line 20: github action is too slow if we're testing all these. Let's just focus on the only version that builds, and let's remove windows as support for windows is too hard anyway :P

.github/workflows/docker.yml
```diff
@@ -1,7 +1,7 @@
 name: ci
 
 on:
-  [push, pull_request]
```
mimoo, line 4: since we tend to push to this repo for pull requests, avoids having redundant jobs

.gitignore
```diff
@@ -9,4 +9,5 @@ _build
 *.install
 *:*
 
-*~
\ No newline at end of file
+*~
+_opam
```
mimoo, line 13: local switches

Dockerfile
```diff
@@ -1,46 +1,39 @@
...
 # library itself later, bind mount your updated source 
 # over this and run the build again.
 COPY . /source
 RUN sudo chown -R opam:opam /source
-VOLUME ["/source"]
+VOLUME /source
 
-# Move to a newer version of OCaml and install dune/jbuilder.
+# Move to a newer version of OCaml and install dune.
 RUN git -C /home/opam/opam-repository pull
```
mimoo, line 28: I wanted to get rid of this line but for some reason it complaines about the version of dune if I do

Dockerfile
```diff
@@ -1,46 +1,39 @@
...
 
 # Pin and install the dependencies.
-RUN eval "$(opam config env)" && \
-    (rm -r .git || true) && \
-    opam pin add -y "/source"
+RUN opam pin add -y --kind path .
 
-# Docker inception
-COPY Dockerfile /
+CMD ["bash" "opam list"]
```
mimoo, line 39: I believe without this we can't run the container interactively

Dockerfile
```diff
@@ -1,46 +1,39 @@
...
+RUN opam upgrade -y
+RUN opam install dune
 
 WORKDIR /source
 
 # Pin and install the dependencies.
-RUN eval "$(opam config env)" && \
-    (rm -r .git || true) && \
-    opam pin add -y "/source"
+RUN opam pin add -y --kind path .
```
mimoo, line 37: I'm guessing the `rm -r .git` was to avoid the pin not working when there are uncommitted changes. We can get around that by just using `--kind path`.

imeckler (approved):
nice!

PR #580, opened 2021-05-24
[sponge] parameterize 'initial_ark' in poseidon permutation 

mimoo:
Our implementation of poseidon has an optional round constant in the very beginning.
We use it thorough the OCaml codebase, but our new params do not use it.
See https://github.com/o1-labs/marlin/blob/master/oracle/src/poseidon.rs#L82

For this reason, we need to be able to disable when needed.

(in addition this gets rid of unused round constants for fp_3)


sponge/sponge.ml
```diff
@@ -171,22 +173,27 @@ module Poseidon (Inputs : Intf.Inputs.Poseidon) = struct
...
+    for i = fst range to snd range do
       (* SBOX -> MDS -> ARK *)
       Array.map_inplace !state ~f:sbox ;
       state := apply_affine_map (mds, round_constants.(i)) !state
     done ;
-    for
-      i = first_half_rounds_full + 1 to first_half_rounds_full + rounds_partial
-    do
+    let range = (snd range + 1, snd range + rounds_partial) in
+    for i = fst range to snd range do
```
mimoo, line 191: I realize that this is ugly... I'm trying to think of a cleaner way to write this but my OCaml-fu is still weak

mimoo:
any idea what these errors are from circleci?

```
[ERROR] The sources of the following couldn't be obtained, aborting:
          - fix.20200131: Bad checksum
          - menhir.20200211: Bad checksum
          - menhirLib.20200211: Bad checksum
          - menhirSdk.20200211: Bad checksum
```

mimoo:
looks like it should have been fixed: https://github.com/ocaml/opam-repository/pull/18529

PR #583, opened 2021-05-26
[CI] circleCI -> github actions

mimoo:
The CircleCI jobs are stuck because of opam / caching.

Since we use both CircleCI and Github actions at the moment, why not just get rid of CircleCI?
Note that this removes caching of opam packages, and so it'll be slower. With the current traffic on this repository it is probably not a huge deal.


mimoo:
since we fixed the CircleCI stuff I'll close this for now

PR #585, opened 2021-05-28
[sponge] parameterize 'initial_ark' in poseidon permutation

mimoo:
Our implementation of poseidon has an optional round constant in the very beginning.
We use it thorough the OCaml codebase, but our new set of parameters (fp_3) does not use it.
(See https://github.com/o1-labs/marlin/blob/master/oracle/src/poseidon.rs#L82)

For this reason, we need to be able to disable that first round constant when needed.

(in addition this PR gets rid of unused round constants for fp_3, and hardens one of our API.)


sponge/sponge.ml
```diff
@@ -171,22 +173,27 @@ module Poseidon (Inputs : Intf.Inputs.Poseidon) = struct
...
+    for i = fst range to snd range do
       (* SBOX -> MDS -> ARK *)
       Array.map_inplace !state ~f:sbox ;
       state := apply_affine_map (mds, round_constants.(i)) !state
     done ;
-    for
-      i = first_half_rounds_full + 1 to first_half_rounds_full + rounds_partial
-    do
+    let range = (snd range + 1, snd range + rounds_partial) in
+    for i = fst range to snd range do
```
mimoo, line 191: I realize that this is ugly... I'm trying to think of a cleaner way to write this but my OCaml-fu is still weak

sponge/sponge.ml
```diff
@@ -209,14 +209,23 @@ module Make_hash (P : Intf.Permutation) = struct
...
+  let to_blocks rate field_elems =
+    let n = Array.length field_elems in
+    let num_blocks = if n = 0 then 1 else (n + rate - 1) / rate in
+    let fill_block block_idx pos =
+      let global_pos = (rate * block_idx) + pos in
+      if global_pos < n then field_elems.(global_pos)
+      else (* padding *) Field.zero
+    in
+    let create_block idx = Array.init rate ~f:(fill_block idx) in
+    Array.init num_blocks ~f:create_block
```
mimoo, line 222: this will work with `field_elems = [| |]` as well now.

sponge/constants.ml
```diff
@@ -3972,190 +3972,6 @@ let params_Pasta_p_3 =
        ; [| "20558520357570880553884246024942640189633857038595234789938692714507322177770"
           ; "4197191416039961928737506336102865351317750605955839230313682916587353308808"
           ; "4737005588414019709913783444410704086294398867618403845170192706541438726511"
-         |]
```
mimoo, line 3975: these extra rounds (99-54) are useless as we only have 54 rounds.

mrmr1993 (approved):
Nice! Looks great

PR #586, opened 2021-06-07
[sponge] fix bugs + test vectors

mimoo:
## What's wrong?

I introduced an off-by-one error here: https://github.com/o1-labs/snarky/pull/585/commits/120fcdae5b697a46b481bd84d435b825dcc4e699#diff-6ece204f52dc5d5f99d41c96cb239b21e8968325f0e6e937aa1eea2be1b1dc57L187

Interestingly, I was thinking about that change these days and the fact that without test vectors I had potentially introduced a bug. Well, looks like I did it. 

## What's in this PR?

So this PR fixes the bug and introduces test vectors (generated by this PR in the marlin repo: https://github.com/o1-labs/marlin/pull/103)

## Does it work?

Yeah, you can run `dune runtest` to run these.

## Potential issues with this PR:

- we're copy/pasting code from the hexstring lib: https://github.com/mimoo/hexstring because hexstring is not compatible with 4.07.1. As we move to a higher version of OCaml I should be able to remove this (cc @psteckler )
- we're also copy/pasting code from Mina's [random_oracle](https://github.com/MinaProtocol/mina/blob/develop/src/lib/random_oracle/random_oracle.ml) as well as the [poseidon.js](https://github.com/MinaProtocol/mina/pull/8913/). I think there's an argument somewhere that the hash part of that code could live here as we're already defining a field and the params in this repo. At the same time, if I understand correctly, we have an OCaml implementation of the field here VS a Rust implementation on the mina repo. Since the latter is more efficient it doesn't make sense to move the code here atm.

### DIFF
--- a/sponge/dune
+++ b/sponge/dune
@@ -1,7 +1,7 @@
 (library
-  (name sponge)
-  (public_name sponge)
-  (preprocess (pps ppx_jane ppx_deriving.eq))
-  (inline_tests)
-  (libraries
-    core_kernel ))
+ (name sponge)
+ (public_name sponge)
+ (preprocess
+  (pps ppx_jane ppx_deriving.eq))
+ (inline_tests)
+ (libraries core_kernel))
--- a/sponge/sponge.ml
+++ b/sponge/sponge.ml
@@ -192,7 +192,8 @@ module Poseidon (Inputs : Intf.Inputs.Poseidon) = struct
       !state.(0) <- sbox !state.(0) ;
       state := apply_affine_map (mds, round_constants.(i)) !state
     done ;
-    let range = (snd range + 1, rounds_full + rounds_partial - 1) in
+    let second_half_rounds_full = rounds_full - first_half_rounds_full in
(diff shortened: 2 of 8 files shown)


mimoo:
any idea why I'm running into this error in CI? The formatter runs clean on my machine:

```
dune exec --root=. app/reformat-snarky/reformat.exe -- -path . -check
File: ./sponge/test_vectors/hash_function.ml has needs to be ocamlformat-ed
Makefile:35: recipe for target 'check-format' failed
make: *** [check-format] Error 1
```

mimoo:
I had to create an `.ocamlformat` file for formatting to work, and after running ocamlformat (`fd '\.ml[i]?$' | xargs -n1 ocamlformat -i`) it seemed like nothing was formatted.

mimoo:
reformat still complains. And now meja complains as well. I see a few routes ahead:

- remove the custom formater and move to ocamlformat
- fix meja tests OR remove meja

wdyt?

mrmr1993 (approved):
You'll need to reformat the test vectors file, but looks good otherwise!

PR #587, opened 2021-06-08
[sponge] fix test

mimoo:
fixes https://github.com/MinaProtocol/mina/pull/8913

### DIFF
--- a/sponge/sponge.ml
+++ b/sponge/sponge.ml
@@ -225,7 +225,7 @@ module Make_hash (P : Intf.Permutation) = struct
   let%test_unit "empty field_elems to_blocks" =
     let blocks = to_blocks 2 [||] in
     assert (Array.length blocks = 1) ;
-    [%test_eq: Field.t] blocks.(0).(0) Field.zero;
+    [%test_eq: Field.t] blocks.(0).(0) Field.zero ;
     [%test_eq: Field.t] blocks.(0).(1) Field.zero
 
   let%test_unit "block" =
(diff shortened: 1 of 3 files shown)


sponge/sponge.ml
```diff
@@ -222,11 +222,12 @@ module Make_hash (P : Intf.Permutation) = struct
     let create_block idx = Array.init rate ~f:(fill_block idx) in
     Array.init num_blocks ~f:create_block
 
-  let%test "empty field_elems to_blocks" =
+  let%test_unit "empty field_elems to_blocks" =
     let blocks = to_blocks 2 [||] in
-    Array.length blocks = 1
-    && blocks.(0).(0) = Field.zero
-    && blocks.(0).(1) = Field.zero
```
mimoo, line 229: do you know why mina was not fine with this test yet CI/my laptop was? I'm not sure how `=` could work here.

mrmr1993 (approved):
I approve these changes for the purposes of merging, but I've tampered with this PR, so waiting on @mimoo to confirm that the new set of changes is fine before we merge.

sponge/sponge.ml
mrmr1993: The mina nonconsensus code uses a field backed by a bigint type that doesn't implement polymorphic comparison. By using an explicit comparison function, we can preserve the original test.
You can trigger the behaviour to confirm the new version works by running `dune runtest --profile=nonconsensus_medium_curves src/nonconsensus/random_oracle/` in the mina repo.

PR #599, opened 2021-10-20
Add a Typ.t to allow deferred public inputs

mrmr1993:
This PR adds `Typ.Internal.delayed`, which converts a `(var, value) Typ.t` to a `(var -> (var, _) Checked.t, unit) Typ.t`, to allow us to allocate circuit variables in the public input, but then fill them later by calling the function returned by `exists`.

This is necessary for SnarkyJS's snapp construction API: we need to 'return' the hash of the parties stack via their public input, but since some parts are determined in the course of running the circuit, we don't have that hash to provide as an input. We can use this instead to pass `unit` as the input, and a small tweak to the pickles library will let us smuggle it back out as an 'output'.

The current implementation is a low-level hack, in the interest of getting this built and working. At some point, as with the generic `if_`, we should tweak the interface of snarky to support these primitives more naturally.


imeckler (approved):

mrmr1993:
Closing, stale.

PR #631, opened 2022-05-19
Compatibility with core v14

Firobe:
This PR makes a single change so that `snarky` can be built with core v14.


mrmr1993 (approved):

Firobe:
I realize I might have to update the CI to v14 for this to build.

PR #632, opened 2022-05-19
Compatility with core v14

Firobe:
Same as https://github.com/o1-labs/snarky/pull/631


mrmr1993 (approved):
Change lgtm, but you'll need to fix CI

Firobe:
Thank you for finishing this, didn't notice before!

PR #650, opened 2022-09-30
Split Checked into Checked and Checked_ast

mrmr1993:
This PR separates the AST implementation of `Checked.t` from the generalized `Checked.Make` functor. This gives a less ambiguous name to `Checked`, which is ambiguous inside `Snark0`.


mimoo (approved):
splendid!

PR #651, opened 2022-09-30
Change Constraint.t from a list to a singleton

mrmr1993:


src/base/run_state.ml
```diff
@@ -46,7 +46,7 @@ type 'field t =
   ; as_prover : bool ref
   ; log_constraint :
       (   ?at_label_boundary:[ `Start | `End ] * string
-       -> ('field Cvar.t, 'field) Constraint.t
+       -> ('field Cvar.t, 'field) Constraint.t option
```
mimoo, line 49: what's the purpose of having a None constraint?

mrmr1993: It allows us to inject a 'marker' either side of a gadget so that we can associate the constraints with that gadget.

PR #657, opened 2022-10-08
Extract the AST interpreter from the rest of the checked runner

mrmr1993:
This PR extracts the logic for the AST interpreter from the rest of the `Checked_runner` module. Most notably, this also restructures the runner so that it will function correctly over any `Checked.t` implementation.

As a side-effect, the `Runtime_error` type can no longer access the full stack of labels, so we have to build them incrementally and generate the error message at the end. This also tweaks the backtrace printing so that we don't elide intermediate backtraces any more (which has long been an ask!)

### DIFF
--- a/src/base/ast_runner.ml
+++ b/src/base/ast_runner.ml
@@ -1,0 +1,97 @@
+open Core_kernel
+
+exception Runtime_error of string list * exn * string
+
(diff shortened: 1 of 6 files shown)


fabrizio-m (approved):
good for me

PR #660, opened 2022-10-27
remove dead code

mimoo:
not used anywhere in Mina either


mrmr1993 (approved):
This was super useful for debugging when we had mixing between functor instances, but I guess it's always still there in history if I need it.

PR #666, opened 2022-10-31
[cvar] move interface of cvar to cvar module

mimoo:
I'm trying to reuse the `Cvar.Make` module type, but it's in `Backend_extended.S`. So I'm moving it to `cvar.ml`. Not sure if this is a good change.

### DIFF
--- a/src/base/cvar.ml
+++ b/src/base/cvar.ml
@@ -1,5 +1,9 @@
 open Core_kernel
 
+(** Main type representing a [Cvar.t]. It represents a small AST over a field ['f]. 
+    We build this AST instead of directly constraining [Cvar.t]s in order to
+    perform some optimizations (mostly merging generic gates).
+    *)
 type 'f t =
   | Constant of 'f
   | Var of int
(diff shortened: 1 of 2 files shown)


src/base/cvar.ml
```diff
@@ -27,17 +32,75 @@ let to_constant_and_terms ~equal ~add ~mul ~zero ~one =
...
+      This operation is considered unsafe as you would normally increment a counter to obtain a variable that doesn't collide with another one.
+    *)
   let of_index v = Var v
 end
 
+(***********************
+ *  SIGS FOR FUNCTORS  *
+ ***********************)
+
+module type cvar_module_type = sig
```
mrmr1993, line 45: Nit: module types normally have a leading uppercase letter (and in fact I'm surprised that the parser accepted this!)

mrmr1993 (approved):

src/base/cvar.ml
mimoo: interesting, looks like the manual uses all uppercases: https://v2.ocaml.org/manual/moduleexamples.html

PR #600, opened 2021-10-26
upload docs via CI

mimoo:
upload HTML docs on github pages via CI

PR #670, opened 2022-11-01
Stop updating docs from PRs!!

mrmr1993:
This fixes an oversight in https://github.com/o1-labs/snarky/pull/600.


mimoo:
ah dang, it was so that I could test the docs while in PR mode :D thanks for fixing

mimoo (approved):

PR #672, opened 2022-11-05
add asm function

mimoo:
this exposes the ASM functionality of kimchi added in https://github.com/o1-labs/proof-systems/pull/831

See https://github.com/MinaProtocol/mina/pull/12125 to see how it's used in tests

### DIFF
--- a/src/base/backend_intf.ml
+++ b/src/base/backend_intf.ml
@@ -16,6 +16,8 @@ module type Constraint_system_intf = sig
 
   val digest : t -> Md5.t
 
+  val get_asm : t -> string
+
   val set_primary_input_size : t -> int -> unit
 
   val set_auxiliary_input_size : t -> int -> unit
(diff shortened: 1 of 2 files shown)


mrmr1993 (changes requested):
There's no need for this to be exposed here. It's better to let this be just the DSL, and update the underlying `Plonk_constraint_system` to expose it

mimoo:
my understanding was that I can't use it from plonk_constraint_system if I don't expose it in the back end interface, because I won't be able to call it. But maybe I'm wrong, I'll have to check your way

mrmr1993:
> The same should have been done with digest as well no?

Yes please. It would be good to delete the whole constraint system interface from the output of `Make`.

mimoo:
Actually, I thought a bit more about it, and if we want to replace snarky with snarky-rs, then any functionalities exposed by plonk_constraint_system should be exposed by snarky instead

mrmr1993:
Not sure I agree. Snarky-rs will run a circuit to create a constraint system, just like snarky does. The methods that the constraint system exposes are completely unrelated to snarky, it's just responsible for creating it.

mimoo:
sg, closing!

PR #673, opened 2022-11-05
Treat *.ml{i} in Linguist as OCaml

kevinji:
Linguist is marking some *.ml{i} files as SML instead of OCaml; this should fix the statistics.


mrmr1993 (approved):

mimoo:
shouldn't this be fixed upstream btw? (https://github.com/github/linguist)

kevinji:
@mimoo I think SML can also use the .ml extension so any language detection is likely going to have some false positives, especially for a simple program that may be valid in both SML and OCaml.

PR #674, opened 2022-11-07
Remove R1CS_constraint_system from snark_intf

mrmr1993:
This PR removes an unused function, and drops the `R1CS_constraint_system` module from the `Snark_intf.*` interfaces.


src/base/snark_intf.ml
```diff
@@ -1100,17 +1094,7 @@ end
...
 (** The imperative interface to Snarky. *)
 module type Run_basic = sig
-  (** The rank-1 constraint system used by this instance. See
-      {!module:Backend_intf.S.R1CS_constraint_system}. *)
-  module R1CS_constraint_system : sig
-    type t
-
-    val digest : t -> Md5.t
-
-    val get_public_input_size : t -> int Core_kernel.Set_once.t
```
mimoo, line 1110: arguably snarky should also be able to return the public input size right? I guess we never have a snarky type that stores the circuit and the constraint system, but I think it'd be a good idea to have these at some point

mimoo (approved):

src/base/snark_intf.ml
mimoo: oh okay, it's removed because it's actually unused. I see that this is not a very useful feature in general

mimoo: oh and there's still `get_primary_input_size`, we should be rename primary -> public

mitschabaude (approved):
makes sense

PR #680, opened 2022-11-14
refactor snark0

fabrizio-m:


src/base/snark0.ml
```diff
@@ -77,387 +80,7 @@ struct
...
-        let ( ||| ) = ( || )
-
-        let not t = Not t
-
-        let any xs = Or xs
-
-        let all xs = And xs
-      end
-    end
+    include Utils.Make (Backend) (Checked) (As_prover_old) (Runner)
```
rbonichon, line 83: Looks like you could use the `As prover` module directly here or am I mistaken?

src/base/utils.ml
```diff
@@ -1,48 +1,474 @@
 open Core_kernel
-
-(** Alias [Cvar] to avoid shadowing. *)
+module Types0 = Types
 module Cvar0 = Cvar
+module Checked_ast = Checked_ast
```
rbonichon, line 4: I'm not sure this re-definition is useful. It even looks like `Checked_ast` is not used in the rest of the file.

src/base/snark0.ml
```diff
@@ -52,6 +52,9 @@ struct
     let field_vars = Array.map fields ~f:(fun x -> Cvar0.Constant x) in
     typ.var_of_fields (field_vars, aux)
 
+  (* to pass to Utils.Make *)
+  module As_prover_old = As_prover
+
```
rbonichon, line 57: Not sure this alias is useful (see next comment)

src/base/utils.ml
```diff
@@ -1,48 +1,474 @@
 open Core_kernel
-
-(** Alias [Cvar] to avoid shadowing. *)
+module Types0 = Types
```
rbonichon, line 2: Searching the file for `Types0` returns no occurrence. Is this alias necessary?

src/base/utils.ml
```diff
@@ -1,48 +1,474 @@
...
-    let%map z = As_prover.read_var z in
-    if Field.equal z Field.zero then (Field.one, Field.zero)
-    else (Field.zero, Field.inv z)
+  module Boolean = Private.Boolean
+
+  let equal, mul, square, div, inv = Private.(equal, mul, square, div, inv)
+
+  let if_, assert_non_zero = Private.(if_, assert_non_zero)
+
+  let equal_vars, equal_constraints = Private.(equal_vars, equal_constraints)
```
rbonichon, line 473: I'm not sure what the aim here is. 
Every function of this module is public for there is no module signature constraining what's exposed by the `Make` functor. 

If we want function aliases defined at toplevel, we could basically do `include Private` and we'd have the same aliases (and way more). 

If we want to limit what's available within the API, we'd like to do it either though defining a module type, or by defining the module interface (the sibling `.mli` file, which does not currently exist)

rbonichon (changes requested):

src/base/snark0.ml
fabrizio-m: Indeed, I was trying to solve another thing and it's no longer necessary.

src/base/utils.ml
fabrizio-m: It probably is not, I copied most of the definitions from `snark0` when I was not sure how much I was going to port, I'll delete the unneeded ones.

src/base/snark0.ml
fabrizio-m: Will remove.

src/base/utils.ml
fabrizio-m: Will remove.

src/base/utils.ml
fabrizio-m: I just wanted to limit while working, I will make an .mli

src/base/utils.ml
```diff
@@ -45,4 +69,392 @@ struct
...
+  (* end *)
+
+  (* module Boolean = Private.Boolean *)
+  module Typ2 = Typ
+
+  (* let equal, mul, square, div, inv = Private.(equal, mul, square, div, inv) *)
+
+  (* let if_, assert_non_zero = Private.(if_, assert_non_zero) *)
+
+  (* let equal_vars, equal_constraints = Private.(equal_vars, equal_constraints) *)
```
rbonichon, line 459: Remove

src/base/utils.ml
```diff
@@ -45,4 +69,392 @@ struct
...
+      let not t = Not t
+
+      let any xs = Or xs
+
+      let all xs = And xs
+    end
+  end
+  (* end *)
+
+  (* module Boolean = Private.Boolean *)
```
rbonichon, line 452: Remove

src/base/utils.ml
```diff
@@ -1,21 +1,45 @@
...
 struct
-  open struct
-    module Cvar = Backend.Cvar
-    module Field = Backend.Field
-    module Constraint = Backend.Constraint
+  open Backend
+
+  open Runners.Make (Backend) (Checked) (As_prover) (Runner)
+
+  (* module Private = struct *)
```
rbonichon, line 24: Remove

src/base/utils.ml
```diff
@@ -45,4 +69,392 @@ struct
...
+      let ( ||| ) = ( || )
+
+      let not t = Not t
+
+      let any xs = Or xs
+
+      let all xs = And xs
+    end
+  end
+  (* end *)
```
rbonichon, line 450: Remove the line

rbonichon (approved):

PR #750, opened 2022-12-08
remove As_prover from Types.Types

mimoo:
builds on top of https://github.com/o1-labs/snarky/pull/686

I'm hoping that this change will help me fully removing checked_ast


mimoo:
doing it as part of https://github.com/o1-labs/snarky/pull/752 as I needed it

PR #761, opened 2022-12-13
Removing checked_ast 3/3

mimoo:
get rids of the AST. Builds on top of https://github.com/o1-labs/snarky/pull/752

mina side PR: https://github.com/MinaProtocol/mina/pull/12360

---

this deletes the snarkydef module, and replaces its usage with direct calls. All it seemed to do was to wrap functions with a `with_label` call.

The change looks like adding these lines in most places:

```ocaml
        let label =
          Stdlib.("compare: " ^ __FILE__ ^ ":" ^ string_of_int __LINE__)
        in
        let%bind () = with_label label (fun _ -> return ()) in
```

This looks like it's correct to me, because `return` will simply start a `Pure ()`, which the `with_label` function will replace with:

```ocaml
Function (fun s -> 
  (* ... *)
  (* evaluate [t] on the state *)
  let new_state_with_label, y = Simple.eval (t ()) state_with_label in
```

and as we can see, `eval` will simply return the real state and `()`:

```ocaml
  let eval (t : ('a, 'f) t) : 'f field Run_state.t -> 'f field Run_state.t * 'a
      =
    match t with Pure a -> fun s -> (s, a) | (* ... *)
```

(I also took the liberty of changing the format slightly to `filename:linenumber` as most editors linkify such format and allow you to click to source on them.)


ppx/dune
```diff
@@ -1,7 +0,0 @@
-(library
```
mrmr1993, line 1: We agreed to remove `snarkydef`, not the whole ppx. Revert.

mrmr1993 (changes requested):

PR #767, opened 2022-12-14
[WIP][easy] remove in_checked_computation (unused)

mimoo:
cc @mitschabaude you said this is unused in snarkyjs right?


mitschabaude:
no I meant it is not necessary and can be easily refactored away :D It might be used in some places still, I can help with removing it when you're doing the Mina PR

PR #768, opened 2022-12-16
[Field.Assert] make lte pure if we're comparing constants

mimoo:
Let's see if this fix a bug in snarkyJS


mitschabaude (approved):
thanks!

PR #772, opened 2022-12-20
[easy] simple cleanings

mimoo:


mrmr1993 (commented):
LGTM, but: does mina compile with this? I think there's still some places that we depend on the toplevel `Typ` implementation (e.g. pickles)

mrmr1993 (approved):

PR #776, opened 2023-01-10
move assert_equals to checked

mimoo:
to fix a bug I mention in the comment here https://github.com/o1-labs/snarky/pull/752#issuecomment-1376600672 

I'll copy/paste the issue:

## The issue

running into some error, that can be reproduce from mina by doing:

```
make snarkyjs
cd src/lib/snarky_js_bindings/test_module 
npm i
node simple-zkapp-mock-apply.js
```

the issue is an `assert_equals` function that does not handle constant variables correctly.

The reason is that it's using what's in `checked.ml` (which is "impure" as it adds a constraint):

```ocaml
  let assert_ ?label c = add_constraint (Constraint.override_label c label)
  
  let assert_equal ?label x y = assert_ (Constraint.equal ?label x y)
```

whereas we want to use the wrapper that's introduced in `utils.ml` (which handles the constant correctly, and keep the logic "pure"):

```ocaml
  let assert_equal ?label x y =
    match (x, y) with
    | Cvar0.Constant x, Cvar0.Constant y ->
        if Field.equal x y then Checked.return ()
        else
          failwithf !"assert_equal: %{sexp: Field.t} != %{sexp: Field.t}" x y ()
    | _ ->
        Checked.assert_equal ?label x y
```

## The solution

I just moved the correct logic to `checked.ml` 

## It works

the following test runs without issue now!

```console
make snarkyjs
cd src/lib/snarky_js_bindings/test_module 
npm i
node simple-zkapp-mock-apply.js
```

Screenshot 2023-01-09 at 6 42 47 PM

### DIFF
--- a/src/base/checked.ml
+++ b/src/base/checked.ml
@@ -1,8 +1,12 @@
 open Core_kernel
 
-module Make
-    (Basic : Checked_intf.Basic)
-    (As_prover : As_prover_intf.Basic with type 'f field := 'f Basic.field) :
+module Make (Field : sig
+  type t [@@deriving sexp]
+
+  val equal : t -> t -> bool
+end)
+(Basic : Checked_intf.Basic with type 'f field = Field.t)
+(As_prover : As_prover_intf.Basic with type 'f field := 'f Basic.field) :
   Checked_intf.S
     with module Types = Basic.Types
     with type 'f field = 'f Basic.field = struct
@@ -72,5 +76,12 @@ module Make
         bind acc ~f:(fun () ->
             add_constraint (Constraint.override_label c label) ) )
 
-  let assert_equal ?label x y = assert_ (Constraint.equal ?label x y)
+  let assert_equal ?label x y =
+    match (x, y) with
+    | Cvar.Constant x, Cvar.Constant y ->
+        if Field.equal x y then return ()
+        else
+          failwithf !"assert_equal: %{sexp: Field.t} != %{sexp: Field.t}" x y ()
+    | _ ->
+        assert_ (Constraint.equal ?label x y)
 end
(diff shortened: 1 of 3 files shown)


src/base/checked.ml
```diff
@@ -72,5 +76,12 @@ module Make
...
 
-  let assert_equal ?label x y = assert_ (Constraint.equal ?label x y)
+  let assert_equal ?label x y =
+    match (x, y) with
+    | Cvar.Constant x, Cvar.Constant y ->
+        if Field.equal x y then return ()
+        else
+          failwithf !"assert_equal: %{sexp: Field.t} != %{sexp: Field.t}" x y ()
+    | _ ->
+        assert_ (Constraint.equal ?label x y)
```
mimoo, line 86: if you're wondering where this leads to, here's the code from plonk_constraint_system:

```ocaml
    | Snarky_backendless.Constraint.Equal (v1, v2) -> (
        (* `red` will create rows in the circuit to encode the variables if need be *)
        let (s1, x1), (s2, x2) = (red v1, red v2) in
        match (x1, x2) with
        | `Var x1, `Var x2 ->
            (* if the scalars match, we can optimize the equality check *)
            if Fp.equal s1 s2 then (
              (* if the scalars are 0, we do nothing *)
              if not (Fp.equal s1 Fp.zero) then
                (* we optimize the equality check by just using the permutation, see https://o1-labs.github.io/proof-systems/snarky/kimchi-backend.html#union-finds *)
                Union_find.union (union_find sys x1) (union_find sys x2) )
            else if (* s1 x1 - s2 x2 = 0
          *)
                    not (Fp.equal s1 s2) then
              add_generic_constraint ~l:x1 ~r:x2
                [| s1; Fp.(negate s2); Fp.zero; Fp.zero; Fp.zero |]
                sys
            else
              add_generic_constraint ~l:x1 ~r:x2
                [| s1; Fp.(negate s2); Fp.zero; Fp.zero; Fp.zero |]
                sys
```

mrmr1993 (approved):

PR #791, opened 2023-02-16
[part1] Generalize Typ.t over Cvar.t

mrmr1993:
This PR generalizes `Typ.t` over `Cvar.t`, by adding a type parameter.

This also makes some tweaks to the functors and module types to make it easier to port this to Mina.

PR #793, opened 2023-02-27
[part2] Pass Cvar through backend

mimoo:
Continue the generalization of cvar happening in https://github.com/o1-labs/snarky/pull/791, now cvar is fully provided by the backend.

Note: This is not mergeable as I two blocks of logic with an exception, these will get fixed in the next PR which will remove run_state.

TODO:

- [x] address the TODO I commented below (see in https://github.com/o1-labs/snarky/pull/798)
- [x] do run_state as well, which will simplify a number of stuff here (https://github.com/o1-labs/snarky/pull/798)
- [ ] then finally do the Mina side (see in https://github.com/o1-labs/snarky/pull/798)


src/base/run_state.ml
```diff
@@ -83,13 +83,14 @@ let get_variable_value { num_inputs; input; aux; _ } : int -> 'field =
...
+  [@@ocaml.warning "-27"]
+(* let v = !next_auxiliary in
+   incr next_auxiliary ; Vector.emplace_back aux x ; Cvar.Unsafe.of_index v *)
 
-let alloc_var { next_auxiliary; _ } () =
-  let v = !next_auxiliary in
-  incr next_auxiliary ; Cvar.Unsafe.of_index v
+let alloc_var { next_auxiliary; _ } () = failwith "todo" [@@ocaml.warning "-27"]
+(* let v = !next_auxiliary in
+   incr next_auxiliary ; Cvar.Unsafe.of_index v *)
```
mimoo, line 93: need to figure this one out, next PR will abstract run_state and remove that hopefully

src/base/run_state.ml
```diff
@@ -83,13 +83,14 @@ let get_variable_value { num_inputs; input; aux; _ } : int -> 'field =
   if i <= num_inputs then Vector.get input (i - 1)
   else Vector.get aux (i - num_inputs - 1)
 
-let store_field_elt { next_auxiliary; aux; _ } x =
-  let v = !next_auxiliary in
-  incr next_auxiliary ; Vector.emplace_back aux x ; Cvar.Unsafe.of_index v
+let store_field_elt { next_auxiliary; aux; _ } x = failwith "todo"
+  [@@ocaml.warning "-27"]
+(* let v = !next_auxiliary in
+   incr next_auxiliary ; Vector.emplace_back aux x ; Cvar.Unsafe.of_index v *)
```
mimoo, line 89: need to figure this one out, next PR will abstract run_state and remove that hopefully

src/base/backend_extended.ml
```diff
@@ -196,27 +162,19 @@ struct
...
   module Constraint = struct
     open Constraint
     include Constraint.T
 
     type 'k with_constraint_args = ?label:string -> 'k
 
-    type t = (Cvar.t, Field.t) Constraint.t [@@deriving sexp]
+    type t = (Cvar.t, Field.t) Constraint.t
+
+    let t_of_sexp _ = failwith "unimplemented"
```
mimoo, line 175: I hope this is OK

src/base/utils.ml
```diff
@@ -23,18 +78,6 @@ struct
...
-    module T = Typ.Make (Checked_intf.Unextend (Checked))
-    include T.T
-
-    type ('var, 'value) t = ('var, 'value, Field.t, Cvar.t) T.t
-
-    let unit : (unit, unit) t = unit ()
-
-    let field : (Cvar.t, Field.t) t = field ()
-  end
-
```
mimoo, line 37: remove reconstructing `Typ` when it's already constructed in snark0 (exact copy/paste here)

src/base/utils.ml
```diff
@@ -1,5 +1,4 @@
 open Core_kernel
```
mimoo, line 1: I removed `utils.mli` because it's a pain to handle the interface when refactoring (and this is all going to disappear anyway)

src/base/backend_extended.ml
mrmr1993: We should delete `sexp` from the interface, can be in a separate PR

src/base/run_state.ml
mrmr1993: You'll probably want to call a `store_field_element` helper on the constraint system

src/base/run_state.ml
mrmr1993: Likewise, `alloc_var` on the constraint system.

mimoo: this is now implemented in rust

src/base/run_state.ml
mimoo: this is now implemented in rust

PR #830, opened 2023-08-24
Avoid costly errors in JS witness generation

mitschabaude:
Instead of throwing JS errors, which are costly because they collect a full stack trace, this changes some core parts of snarky to throw strings instead. Throwing strings is very fast, and this change leads to a huge speedup during witrness generation, see https://github.com/o1-labs/o1js/pull/1092

### DIFF
--- a/src/base/checked_runner.ml
+++ b/src/base/checked_runner.ml
@@ -256,8 +256,11 @@ struct
           let old = Run_state.as_prover s in
           Run_state.set_as_prover s true ;
           let value =
-            As_prover.Provider.run p (Run_state.stack s) (get_value s)
-              (Run_state.handler s)
+            As_prover.Provider.run p (get_value s) (Run_state.handler s)
+            |> Option.value_exn
+                 ~message:
+                   ( "Unhandled request: "
+                   ^ Core_kernel.String.concat ~sep:"\n" (Run_state.stack s) )
           in
           Run_state.set_as_prover s old ;
           let var =
--- a/src/base/checked_runner.ml
+++ b/src/base/checked_runner.ml
@@ -256,11 +256,15 @@ struct
           let old = Run_state.as_prover s in
           Run_state.set_as_prover s true ;
           let value =
-            As_prover.Provider.run p (get_value s) (Run_state.handler s)
-            |> Option.value_exn
-                 ~message:
-                   ( "Unhandled request: "
-                   ^ Core_kernel.String.concat ~sep:"\n" (Run_state.stack s) )
+            match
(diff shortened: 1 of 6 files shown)


src/base/constraint.ml
```diff
@@ -70,7 +70,11 @@ end = struct
 
     let to_basic x = T x
 
-    let of_basic = function T x -> x | _ -> failwith "different constructor"
+    let of_basic = function
+      | T x ->
+          x
+      | _ ->
+          raise (Failure "different constructor")
```
mitschabaude, line 77: I'm just replacing `failwith s` with `raise (Failure s)`. This makes a difference because in JS, we use a custom override of the `failwith` function, to have it throw JS errors instead of strings. This is very important in a couple of circumstances where the error implies an actual failure condition. Here, the error is just used for control flow and always caught a level earlier, so we don't need a pretty error.

Trivo25 (approved):

src/base/constraint.ml
mrmr1993: If you don't want to pay the cost of the exception here, can you change the code above it to use an `val of_basic : ('v, 'f) basic -> ('v, 'f) t option` instead.

src/base/request.ml
```diff
@@ -28,9 +28,10 @@ module Handler = struct
    fun stack0 label_stack req0 ->
     let rec go req = function
       | [] ->
-          failwith
-            ( "Unhandled request: "
-            ^ Core_kernel.String.concat ~sep:"\n" label_stack )
+          raise
```
mrmr1993, line 31: As above, this should probably return an `option`, and the caller should handle the error by constructing an appropriate error message instead. This also has the benefit of never constructing this string if it isn't needed.

mrmr1993 (changes requested):

src/base/request.ml
mitschabaude: good point! will change

src/base/checked_runner.ml
```diff
@@ -256,8 +256,11 @@ struct
           let old = Run_state.as_prover s in
           Run_state.set_as_prover s true ;
           let value =
-            As_prover.Provider.run p (Run_state.stack s) (get_value s)
-              (Run_state.handler s)
+            As_prover.Provider.run p (get_value s) (Run_state.handler s)
+            |> Option.value_exn
```
mrmr1993, line 260: If you do this in a match statement, then we'll only compute the error message when there's actually an error.

mrmr1993 (approved):

PR #836, opened 2024-02-11
Add callback-oriented versions of `constraint_system` and `generate_witness`

mrmr1993:
This PR refactors the internals of snarky to expose a callback-oriented version of the functions `constraint_system` and `generate_witness`. As the core entrypoints used by pickles, this allows us to use these callbacks to handle async witness generation.

This PR is structured as a series of refactoring commits, plus a final commit that implements the new functions. Each refactoring is deliberately structured so that there is no behavioural change, except for a few small performance optimisations as a result of removing duplicated/redundant computations.


src/base/snark0.ml
```diff
@@ -1287,6 +1287,53 @@ module Run = struct
...
+    type ('input_var, 'return_var, 'result) manual_callbacks =
+      { run_circuit : 'a. ('input_var -> unit -> 'a) -> 'a
+      ; finish_computation : 'return_var -> 'result
+      }
+
+    let constraint_system_manual ~input_typ ~return_typ =
+      let builder =
+        Run.Constraint_system_builder.build ~input_typ ~return_typ
+      in
+      (* FIXME: This behaves badly with exceptions. *)
```
mitschabaude, line 1299: see my review on a way to improve this

mitschabaude (approved):
to let callers recover from errors, we could add a third callback, which just resets the state to what it was:
```ocaml
type ('input_var, 'return_var, 'result) manual_callbacks =
      { run_circuit : 'a. ('input_var -> unit -> 'a) -> 'a
      ; finish_computation : 'return_var -> 'result
      ; reset_state : unit -> unit
      }
```

and you'd use it like this pseudo-code

```ts
let builder = constraint_system_manual();
try {
  await build.run_circuit(input);
  builder.finish_computation();
} catch (err) {
  builder.reset_state();
}
```

since its optional to use use `reset_state` and handle errors, this would be an easy change.

still approving though because it seems that the present version is good enough (passes all tests across the stack, even those that intentionally create errors in circuits - probably because snarky's current state, however broken, never affects subsequent circuit runs, because states are just nested)

querolita (approved):

PR #839, opened 2024-02-28
Support async witnessing

mitschabaude:
see https://github.com/o1-labs/o1js-bindings/pull/253


src/base/snark_intf.ml
```diff
@@ -1122,6 +1122,9 @@ module type Run_basic = sig
   (** The finite field over which the R1CS operates. *)
   type field
 
+  (* a reference to the internal Run_state.t *)
+  val state : field Run_state.t ref
```
rbonichon, line 1126: Would really prefer to keep this internal. And expose a pair of `get_state: unit -> field Run_state.t`, `set_state: field Run_state.t -> unit` functions. This permits any sort of checking to be implemented in there.

mitschabaude: oh you mean enables us to add checks on any state before it is used for example. that makes sense

rbonichon: In no particular order:
1. easier to grep for `set_state` than `:=` 😄 
2. you can check for the value you set, or transform it as you wish, if you need to respect some internal invariants.

rbonichon (approved):

src/base/snark0.ml
```diff
@@ -724,6 +724,10 @@ module Run = struct
            ~eval_constraints:false ~num_inputs:0 ~next_auxiliary:(ref 0)
            ~with_witness:false ~stack:[] ~is_running:false () )
 
+    let get_state () = !state
```
mrmr1993, line 727: I would much prefer not to spew the internals here. Can you use the callback pattern, like [here](https://github.com/o1-labs/snarky/pull/836), instead?

mrmr1993 (changes requested):

src/base/snark_intf.ml
```diff
@@ -1402,6 +1402,9 @@ module type Run_basic = sig
        , Proof_inputs.t * 'return_value )
        manual_callbacks
 
+  (* Callback, low-level version of [as_prover] and [exists]. *)
+  val as_prover_manual : int -> field array option -> Field.t array
```
mrmr1993, line 1406: Style nit: it's better to use `Staged.stage` so that you get an `int -> (field array option -> Field.t array) Staged.t`, which makes it obvious that you shouldn't pass both arguments at the same time.

mrmr1993 (approved):

PR #842, opened 2024-03-12
Be aggressive about checking state consistency

mrmr1993:
This PR adds IDs to the snarky run state, and uses them to check that the state wasn't clobbered by some concurrent circuit execution.


mitschabaude (approved):
very important change, only downside is that the failwiths don't seem to play nicely with alcotest (although that feels more like an alcotest problem to me)

PR #850, opened 2024-12-11
Remove unused code, simplify `Typ.Internal.ref` to `Typ.prover_value`

mrmr1993:
This PR eliminates some unused code. This also replaces the complicated mutable `As_prover.Ref.t` with an immutable `prover_value` opaque type, reflecting its general usage.


src/base/typ.ml
```diff
@@ -145,24 +132,19 @@ module Make (Checked : Checked_monad) = struct
         ; check = (fun _ -> Checked.return ())
         }
 
-    module Internal = struct
-      let snarkless value : _ t =
+    include struct
```
dannywillems, line 135: image

dannywillems (approved):

dannywillems:
I would trust the fact it compiles. I don't have the full context of the usage of the type in the module.

PR #854, opened 2024-12-16
Remove polymorphism from `As_prover.t`, hoist `Typ.t`

mrmr1993:
This PR bundles a few changes. This can be broken into smaller pieces by commit if needed.

The goal of this PR is to remove the generalisation over `Field.t` from `Checked.t` and `As_prover.t`, allowing us to hoist `Typ.t` from the global implementation in `Types`. The result is a unique `Typ.t` type for each instantiation of snarky, allowing us to generate it from any checked backend (and in particular from a rust backend).

To achieve this, we
* create a new helper functor `Types.Make_types`
* remove the free field argument from `As_prover.t`
* replace uses of the concretised types with those from the shared `Types` module signature
* remove some unused or now-unnecessary type and functor machinery.

The recommended way to review this PR is to 'ignore whitespace' in the GitHub UI (or equivalently using `git diff --ignore-all-space`), to ignore the indentation changes resulting from functorisation.

### DIFF
--- a/src/base/as_prover0.ml
+++ b/src/base/as_prover0.ml
@@ -1,10 +1,13 @@
 open Core_kernel
 
-module Make
-    (Types : Types.Types
-               with type ('a, 'f) As_prover.t = ('f Cvar.t -> 'f) -> 'a) =
+module Make (Backend : sig
+  module Field : sig
+    type t
+  end
+end)
+(Types : Types.Types with type ('a, 'f) As_prover.t = ('f Cvar.t -> 'f) -> 'a) =
 struct
-  type ('a, 'f) t = ('a, 'f) Types.As_prover.t
+  type 'a t = ('a, Backend.Field.t) Types.As_prover.t
 
   let map t ~f tbl =
     let x = t tbl in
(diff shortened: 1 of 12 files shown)


src/base/as_prover0.ml
```diff
@@ -56,29 +59,30 @@ module Provider = struct
...
-        | None | (exception _) ->
-            Some (run c tbl)
-        | x ->
-            x )
-end
-
-module Handle = struct
-  let value (t : ('var, 'value) Handle.t) : ('value, 'field) t =
-   fun _ -> Option.value_exn t.value
+    type nonrec ('a, 'f) t = ('a, 'f) Types.Provider.t
```
dannywillems: It was confusing first that the type changes. But ok :).

src/base/typ.ml
```diff
@@ -1,46 +1,5 @@
 open Core_kernel
 
-module Data_spec0 = struct
```
dannywillems, line 4: Yay.

src/base/types.ml
```diff
@@ -58,8 +58,6 @@ module Typ = struct
           ('var, 'value, 'aux, 'field, 'checked) typ'
           -> ('var, 'value, 'field, 'checked) typ
   end
-
-  include T
```
dannywillems, line 62: Yay.

src/base/checked_runner.ml
```diff
@@ -15,7 +15,41 @@ module Simple_types = struct
   end
 
   module Typ = struct
-    include Types.Typ.T
```
dannywillems: Yay.

src/base/checked_intf.ml
```diff
@@ -112,33 +112,7 @@ module type S = sig
...
-    S
-      with module Types := Types
-      with type field := field
-       and type 'a t := ('a, field) Types.Checked.t
+  include S
 
   val run : 'a t -> field Run_state.t -> field Run_state.t * 'a
 end
-
-module Unextend
```
dannywillems: Yay.

dannywillems (approved):
Reviewed commit by commit. Makes sense so far. Curious to see the final result.

PR #855, opened 2024-12-16
Why use `https://` instead of `git@`?**

sky-coderay:
Using `https://` instead of `git@` ensures the command works for everyone. The `git@` format requires configured SSH keys, while `https://` works out-of-the-box, making it more accessible for beginners.


sky-coderay:
@dannywillems Please take a look when you have time

PR #858, opened 2025-01-06
Abstract over `Run_state`, include it in the snarky backend

mrmr1993:
This PR abstracts over `Run_state`, accepting it as a parameter from the snarky backend. This allows us to pass through a rust implementation in place of the default OCaml implementation.


src/base/run_state_intf.mli
```diff
@@ -0,0 +1,89 @@
+module Vector : sig
```
dannywillems, line 1: Wondering if making an external library for vectors (and the other types we use in Pickles) would make sense. Could be nice for the OCaml community too.

dannywillems: Not blocking, but I don't understand why we have the nested module `Vector` with the methods, and externally we do have only the type, both types being equal.

src/base/checked_runner.ml
```diff
@@ -341,12 +349,14 @@ struct
           Checked_intf.Basic
             with module Types := Types
             with type field := Checked_runner.field
+             and type run_state := run_state
```
dannywillems, line 352: Not blocking: wondering if we could not remove some type aliases above to simplify the interface. `type state = run_state` for instance. Only looking at the lines above, it might not make sense as used for something else.

src/base/runners.ml
```diff
@@ -215,7 +217,7 @@ struct
           Runner.State.make ~num_inputs ~input ~next_auxiliary ~aux ~system
             ~with_witness:false ()
         in
-        let id = Run_state.id state in
+        let id = Backend.Run_state.id state in
```
dannywillems, line 220: Nit: this doesn't seem to be useful (compiles without on my side), as the module is opened L21. Or, are you trying to remove the statements `open` :face_holding_back_tears:?

src/base/snark0.ml
```diff
@@ -714,7 +715,7 @@ module Run = struct
 
   module Make_basic (Backend : Backend_intf.S) = struct
     module Snark = Make (Backend)
-    open Run_state
+    open Backend.Run_state
```
dannywillems, line 718: It seems we can remove this open by changing L1564. See https://github.com/o1-labs/snarky/commit/242d5b55a169f93c1540ec75a3bfc90c81c4cbac.

PR #859, opened 2025-01-06
Abstract over the specific `Constraint.t` type used by the backing constraint system

mrmr1993:
This PR builds on https://github.com/o1-labs/snarky/pull/858, delivering on a large part of the promise of previous iterations: this allows us to define a constraint-system-specific constraint enum instead of overloading the snarky one. In particular, this gives the ability to have safe rust FFI types for constraint system generation, by only allowing the specific constraints that it supports.

As part of the pre-work that built up to this, we have also ensured that a `Typ.t` for any particular constraint definition is compatible with only the constraint systems that support it, ensuring that we get better type safety for OCaml / JS too.

### DIFF
--- a/src/base/backend_extended.ml
+++ b/src/base/backend_extended.ml
@@ -59,8 +59,6 @@ module type S = sig
     val to_constant : t -> Field.t option
   end
 
-  module R1CS_constraint_system : Constraint_system.S with module Field := Field
-
   module Constraint : sig
     type t = (Cvar.t, Field.t) Constraint.t [@@deriving sexp]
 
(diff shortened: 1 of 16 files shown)


dannywillems:
Changing the base branch to master. It seems it is not doing automatically. GH having some issues?

src/base/backend_extended.ml
```diff
@@ -64,22 +64,15 @@ module type S = sig
   module Constraint : sig
     type t = (Cvar.t, Field.t) Constraint.t [@@deriving sexp]
 
-    type 'k with_constraint_args = ?label:string -> 'k
+    val boolean : Cvar.t -> t
```
dannywillems, line 65: Ah, I thought we were using it everywhere in Pickles.

dannywillems (approved):

dannywillems:
Context clearer now. We want to avoid mixing constraints from different backends too.

PR #868, opened 2025-04-24
refactors draft

martyall:
This is the final product to see how far I can get, it will be broken up into much smaller prs. 

I have been going through snarky code in depth recently and have noticed that it could use some love. The primary symptom of this is that often times closely related code seems to be really spread out, resulting in an incredible number of files being open in my editor to just read basic stuff. 

The ultimate goal is to (without changing the meaning or interface):
  - delete unused code
  - remove unnecessary indirection
  - clean up the module hierarchy
  - move out auxillary code (e.g. `monad_lib`)
  - create accompanying mli files


martyall:
image

posting this before and after as a checkpoint

PR #870, opened 2025-04-25
As_prover refactor

martyall:
Currently some modules use the `as_prover0.ml` implementation directly where they might as well just use the `as_prover_intf.ml` module (e.g. `checked_runner.ml` and `merkle.ml`. 

This PR makes is so that `snark0.ml` is the only place which uses the concrete `as_prover` for consistency sake. This follows the same usage pattern for all of the other `my_module` vs `my_module_intf` in this lib

I also renamed from `as_prover0.ml` to `as_prover.ml` because why the `0` if there is no other? Or maybe there's another reason for this / ocaml convention I don't know about?


glyh (approved):

dannywillems:
> I also renamed from as_prover0.ml to as_prover.ml because why the 0 if there is no other? Or maybe there's another reason for this / ocaml convention I don't know about?

The naming convention `As_prover0`, `Snark0`, etc. comes from JaneStreet. It is a widely used pattern in their codebase, used to maintain compatibility and being able to keep different versions over time.
I do not have a strong opinion in favor of it, but you can see that this pattern is also used in other places, where, I guess, Izaak went into :).

dannywillems (approved):
I went through each commit. LGTM.

Issue #867, opened 2025-04-23
Use libs for Monad_let and Monad_sequence


martyall:
The modules `Monad_let` and `Monad_sequence` are defining very basic things that probably don't belong in this library (e.g. in haskell these are part of any prelude). I noticed that they are not used elsewhere in the mina project, so I would guess they are more internal and should probably not be exported by the top level `Snarky` module.

`Hlist` feels similar

PR #871, opened 2025-04-26
Move monad code into its own library

martyall:
partially addresses #867 

Move all of the generic monad code into a library `snarky_monad_lib`, similar to what we do with `tuple_lib`, `fold_lib`, etc. Honestly I would have preferred the name `monad_lib`, but when building the mina project (which defines it's own `monad_lib` library)  I was getting the error

```
Error: The files src/lib/currency/.currency.objs/byte/currency.cmi
       and src/lib/monad_lib/.monad_lib.objs/byte/monad_lib.cmi
       make inconsistent assumptions over interface Monad_lib
```

which makes me think that ocaml has some kind of globally unique constraint on module names. If this is not the case (which I hope it isn't), I can rename it


glyh (approved):
I don't really know these low level details of Dune to understand why there's name clashing, sorry :(

Codes LGTM :)

snarky_monad_lib/dune
```diff
@@ -0,0 +1,9 @@
+(library
+ (name snarky_monad_lib)
```
dannywillems, line 2: Note that _lib is not necessary, and I feel it is a bit redundant as we know it is a library. API users would have to add this suffix, which might not sound natural. Although, I think it's mostly to follow the naming convention in Snarky you're doing this?

snarky_monad_lib/dune
```diff
@@ -0,0 +1,9 @@
...
+ (public_name snarky.monad_lib)
+ (library_flags -linkall)
+ (inline_tests)
```
dannywillems, line 5: Are there tests? :o

snarky_monad_lib/dune
```diff
@@ -0,0 +1,9 @@
+(library
+ (name snarky_monad_lib)
+ (public_name snarky.monad_lib)
+ (library_flags -linkall)
```
dannywillems, line 4: I would check if it is necessary with the recent dune versions. I know it's everywhere in the codebase, but not sure it is required.

snarky_monad_lib/dune
```diff
@@ -0,0 +1,9 @@
...
+ (inline_tests)
+ (libraries core_kernel)
+ (preprocess
+  (pps ppx_jane ppx_deriving.eq bisect_ppx -- --conditional))
```
dannywillems, line 8: Interesting for --conditional, I did not know it. Where does it come from?

dannywillems (commented):
LGTM, I left a few follow up comments.

dannywillems (approved):

snarky_monad_lib/dune
martyall: Yeah I just followed the naming convention from `tuple_lib`. And in fact you will find your dune file comments apply equally to there because I took that dune file as a template.

PR #872, opened 2025-04-26
Remove unused constraint module

martyall:
This must be some legacy module from another time. It's not being used -- not internally in snarky, nor in mina


dannywillems (approved):
Interesting.

PR #873, opened 2025-04-27
remove unused bin_prot_io module

martyall:
This module isn't being used in this library or in the mina project, and doesn't really belong in this lib -- it's generic code that doesn't pass the [Fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold#:~:text=The%20Fairbairn%20threshold%20is%20the,common%20use%20several%20years%20ago.)


boray:
Can you remove bin_prot_io mentions from website as well?

dannywillems (approved):

PR #882, opened 2025-04-29
Add Fermat test

martyall:


src/tests/fermat.ml
```diff
@@ -0,0 +1,82 @@
...
+    let typ = Impl.Field.typ in
+    Impl.Typ.transport typ
+      ~there:(fun (Solve_cube_root x) -> x)
+      ~back:(fun x -> Solve_cube_root x)
+
+  type _ Snarky_backendless.Request.t +=
+    | Cube_root : cube_root Snarky_backendless.Request.t
+
+  (* TODO: Why isn't this defined in the field interface ? *)
+  let pow x n =
```
dannywillems: Good question.

src/tests/backend.ml
```diff
@@ -40,7 +40,7 @@ module Serialize = struct
 end
 
 module Fp (P : sig
-  val order : Bignum_bigint.t
+  val characteristic : Bignum_bigint.t
```
dannywillems, line 43: I would keep order. It is not equivalent when we have an extension.

martyall: FWIW my goal here is to just make a prime field, the use of the Bigint type as the implementation wouldn't even allow for anything else. Also want to note that `Bigint_intf` requires a `from_field` method which also doesn't make any sense for an extension. 

The fact that `Bigint_intf` relies on a field type and not the other way around is already a weird smell

martyall:
closing in favor of #884

PR #885, opened 2025-05-01
Add direnv support

glyh:
The standard workflow with OCaml would create the switch on the git root. This envrc profile add the command needed to source the switch.


dannywillems (changes requested):
I would be against this change as not everyone uses a local switch.

dannywillems:
Closing based on my comment above.

PR #887, opened 2025-05-06
Bump up docusaurus

dannywillems:
It won't solve the fact the doc is broken when deployed.
Docusaurus in NPM is old, 4 years old, see https://www.npmjs.com/package/docusaurus.


dannywillems:
Closing. Missing `make website` step, and it is indeed broken because of Highlight.js EOL

PR #900, opened 2025-05-14
Document sponge

boray:
This PR adds comments to `sponge/sponge.mli`.

### DIFF
--- a/sponge/sponge.mli
+++ b/sponge/sponge.mli
@@ -166,7 +166,7 @@ module Bit_sponge : sig
       Intf.Sponge
         with module State := State
          and module Field := Field
-         (* The digest is a list of booleans of a specified length. *)
+        (* The digest is a list of booleans of a specified length. *)
          and type digest := length:int -> Bool.t list
          and type input := Input.t
          and type t = (S.t, Bool.t) t
--- a/sponge/sponge.mli
+++ b/sponge/sponge.mli
@@ -171,9 +171,7 @@ module Bit_sponge : sig
          and type input := Input.t
          and type t = (S.t, Bool.t) t
 
-    (** Squeezes a field element from the sponge, bypassing the bit conversion.
-        This is useful when the caller needs a field element.
-    *)
+    (** Squeezes a field element from the sponge, bypassing the bit conversion. *)
     val squeeze_field : t -> Field.t
   end
 end
--- a/sponge/sponge.mli
+++ b/sponge/sponge.mli
@@ -1,4 +1,6 @@
-(* Defines the interfaces for sponge constructions. *)
+(* Defines the interfaces for sponge constructions.
+   https://en.wikipedia.org/wiki/Sponge_function
+*)
 module Intf = Intf
 
 (** Parameters for the sponge construction. *)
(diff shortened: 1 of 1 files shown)


45930 (commented):
I don't think these comments are very helpful.  In some cases, they are just too simple e.g. 

```ocaml
(* Defines the interfaces for sponge constructions. *)
module Intf = Intf

(** Represents the internal state of the sponge. *)
module State : sig
```

And in other cases, it references concepts that may not be obvious without explaining them, e.g.

```ocaml
(** Functor to create a hash function from a permutation. *)
module Make_hash (P : Intf.Permutation) :
  Intf.Hash with module State := State and module Field := P.Field
```

Basically, if you don't know what this code does, I don't think the comments help at all, and if you do know what the code does, the comments are unnecessary.  IMO the kinds of things I would want to see documented here would be: What is a sponge (maybe link to Wikipedia)?, how do I instantiate a sponge?, What are the implications of using a Field sponge or a bit sponge?, What are the rescue permutation and the poseidon permutation (where are these defined)? - How do I contextualize this code within the project?

sponge/sponge.mli
```diff
@@ -1,49 +1,79 @@
...
   type 'a t = 'a array
 
+  (** Maps over the state. *)
   val map : 'a t -> f:('a -> 'b) -> 'b t
 end
 
+(** Instantiation of the Rescue permutation.
+    [Inputs] defines the field and other parameters for the Rescue permutation.
+*)
 module Rescue (Inputs : Intf.Inputs.Rescue) :
```
dannywillems, line 35: FWIW: Rescue has been introduced in https://eprint.iacr.org/2019/426. It is an instantiation of the Marvellous construction.

boray:
> I don't think these comments are very helpful. In some cases, they are just too simple e.g.
> 
> ```ocaml
> (* Defines the interfaces for sponge constructions. *)
> module Intf = Intf
> 
> (** Represents the internal state of the sponge. *)
> module State : sig
> ```
> 
> And in other cases, it references concepts that may not be obvious without explaining them, e.g.
> 
> ```ocaml
> (** Functor to create a hash function from a permutation. *)
> module Make_hash (P : Intf.Permutation) :
>   Intf.Hash with module State := State and module Field := P.Field
> ```
> 
> Basically, if you don't know what this code does, I don't think the comments help at all, and if you do know what the code does, the comments are unnecessary. IMO the kinds of things I would want to see documented here would be: What is a sponge (maybe link to Wikipedia)?, how do I instantiate a sponge?, What are the implications of using a Field sponge or a bit sponge?, What are the rescue permutation and the poseidon permutation (where are these defined)? - How do I contextualize this code within the project?

I thought the target audience as non-ocaml o1labs devs who is trying to understand mina/snarky codebase. Thus, I assumed the dev knows sponge construction, functors, field/bit  etc. sponge as a priori. 

My current methodology is generating comments with an LLM then understanding the code and correcting or refining the comments. I can change this if it's better to widen the target audience.

sponge/sponge.mli
boray: It's already in the sponge.ml comments. What do you think about the documentation/comments scattered in two files?

dannywillems (changes requested):
The code does not give actual valuable information. Writing documentation is not the best way to facilitate usage of a library. Examples and tests are the best way to achieve such a goal.

PR #903, opened 2025-06-01
Make global state an option

L-as:
The PR exists merely to show why the fork exists.


L-as:
(though it might still be useful to merge one day)

PR #912, opened 2025-11-29
Bump to (OCaml4.14 + Core.v0.16) / (OCaml5.2 + Core.v0.17)

glyh:
This is part of the bigger plan bumping Mina to OCaml 5.2.1 with Corev0.17.1

### DIFF
--- a/group_map/group_map.ml
+++ b/group_map/group_map.ml
@@ -57,7 +57,7 @@ WB19: Riad S. Wahby and Dan Boneh, Fast and simple constant-time hashing to the 
                    f(u + y^2)·(y^2 + uv + v^2 + ay)/y )
 *)
 
-open Core_kernel
+open Core
 module Field_intf = Field_intf
 module Bw19 = Bw19
 
@@ -91,11 +91,13 @@ module type S = sig
   end
 
   module Make
-      (Constant : Field_intf.S) (F : sig
+      (Constant : Field_intf.S)
+      (F : sig
         include Field_intf.S
 
         val constant : Constant.t -> t
-      end) (Params : sig
+      end)
+      (Params : sig
         val params : Constant.t Params.t
       end) : sig
     val potential_xs : F.t -> F.t * F.t * F.t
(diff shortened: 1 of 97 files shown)


group_map/group_map.ml
```diff
@@ -91,11 +91,13 @@ module type S = sig
   end
 
   module Make
-      (Constant : Field_intf.S) (F : sig
```
glyh, line 94: These are auto format

dannywillems:
FWIW: OCaml 5.2.x is already 2 years old, see https://ocaml.org/releases. OCaml 5.5.0 brings a set of interesting features, and there have been other improvements since 5.2.x.

glyh:
We're bumping up step by step. But I'd love to bump to 5.5.0 if there's no obvious blockers as well -- it just need to be investigated.

Thanks for the recommendation @dannywillems  :)

dkijania (dismissed):

dkijania:
@glyh question if existing tests are enough to cover this bump?

glyh:
@dkijania Good question — I went through the full diff to answer it. I believe the existing tests are sufficient for this bump, for two reasons:

**1. The changes are mechanical, not semantic.** The ~430 changed lines fall entirely into these buckets:
- `Core_kernel` → `Core` renames (Core v0.17 merged Core_kernel into Core) and `Caml.*` → `Stdlib.*`;
- `Sys.getenv_opt` → Core's `Sys.getenv` (same `string option` semantics) and `Sys.getcwd` → `Sys_unix.getcwd`;
- ocamlformat 0.28 reformatting (functor argument indentation etc.);
- dune/opam constraint bumps, CI and Dockerfile updates.

There are no algorithmic or data-structure changes, so what needs coverage is "does everything still compile and behave the same", not new logic.

**2. What the suite exercises is the right regression surface for this kind of change.** `dune runtest` runs: the `src/tests` Alcotest suite (Checked/Typ/As_prover machinery over the pure-OCaml test backend + the Fermat example), inline `let%test`s across `snark0`, `utils`, `sponge`, `group_map` and `snarkette/fields`, the fold_lib/interval_union suites, and the sponge Poseidon test vectors — the last one being a good end-to-end check that the Core bump didn't perturb hashing.

Two caveats, one of which is now addressed:
- CI originally only built 5.2.1 after this PR, while the opam constraints still accept Core v0.16. I've added a symmetric CI matrix (c76de53a) that pins the toolchain pairs explicitly — OCaml 5.2.1 + core.v0.17.2 and OCaml 4.14.2 + core.v0.16.2 — so both supported combinations are built and tested on every PR.
- The ultimate acceptance test is building Mina against this branch, since snarky's own tests run on a dummy backend rather than the real Kimchi integration. That part is outside this repo's suite.

